PR #79, opened 2014-12-27
Added support for 8 digit codes

ercpe:
The default number of digits remains at 6 as the Google Authenticator app does not support 8 digit codes (https://code.google.com/p/google-authenticator/issues/detail?id=327).

However, 8 digit codes can be used with other authenticator apps like the Yubico Authenticator.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -53,6 +53,13 @@ General Settings
   
   For more QR factories that are available see python-qrcode_.
 
+``TWO_FACTOR_TOTP_DIGITS`` (default: ``6``)
+  The number of digits to use for TOTP tokens. Can be set to 6 or 8.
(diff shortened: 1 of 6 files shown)


Bouke:
Your change looks like a nice feature. The change to the settings + documentation suggest that token length is applied for all token mechanisms. However, it appears that it only affects the token generators. I think the change should also be applied for tokens delivered by phone call and text message. Could you have a look into those as well?

ercpe:
I've added the support for 8 digit codes to the Twilio backend implementation. Please not that i haven't tested it with a real device.

Bouke:
Thanks for contributing! :+1:

ercpe:
Great. Is there an ETA for the 1.1.0 release? :)

Bouke:
I've just published 1.1.0 :shipit:

PR #90, opened 2015-04-03
Make compatible with Django 1.8

csghormley:
Patch imports and setup to use django-formtools instead of the obsolete contrib package.

This fixes issue https://github.com/Bouke/django-two-factor-auth/issues/89

Passed all tests in my environment.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,4 +1,4 @@
 [bumpversion]
-current_version = 1.1.1
+current_version = 1.1.1a
 files = setup.py docs/conf.py
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,7 +53,7 @@ copyright = '2014, Bouke Haarsma'
 #
 
 # The full version, including alpha/beta/rc tags.
-release = '1.1.1'
+release = '1.1.1a'
 
 # The short X.Y version.
 version = '.'.join(release.split('.')[0:2])
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='django-two-factor-auth',
-    version='1.1.1',
+    version='1.1.1a',
     description='Complete Two-Factor Authentication for Django',
     long_description=open('README.rst').read(),
     author='Bouke Haarsma',
@@ -12,9 +12,10 @@ setup(
     license='MIT',
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
-        'Django>=1.4.2,<1.8',
+        'Django>=1.4.2,<1.9',
         'django_otp>=0.2.0,<0.3.0',
         'qrcode>=4.0.0,<5.0.0',
+        'django_formtools==1.0',
     ],
     include_package_data=True,
     classifiers=[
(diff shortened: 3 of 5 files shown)


Bouke:
Including django_formtools as a dependency will not work for Django versions <1.7, as django_formtools requires >=1.7. What would be a portable way to continue support for older Django versions?

csghormley:
I can modify the compat.py test:

if django.VERSION[:2] >= (1, 8):
   ...use formtools.*
elif django.VERSION[:2] >= (1, 6):
   ...use django.contrib.formtools.*

One thing I don't know how to do is craft install_requires to 
conditionally require django_formtools.

troygrosfield:
@csghormley, look at the **Requirements in setup.py** portion of this stackoverflow answer:

- http://stackoverflow.com/a/9418401/1007358

So maybe something along the lines of:

```
# setup.py
from setuptools import setup
install_requires = ['somelib1', 'somelib2']

try:
    # try to detect django and django.contrib.formtools?
except ImportError:
    # if django exists, and django.contrib.formtools isn't found
    # add the django_formtools to install_requires
    install_requires.append('django_formtools')

setup(
    ...
    install_requires=install_requires,
    ...
)
```
Just a thought...

troygrosfield:
Second thought, why not just always include ``django_formtools`` in ``install_requires``?  It simply wouldn't be used for users on django < 1.8 and would work for django 1.8 users which would still make it backward compatible for older django versions.

Bouke:
Thanks for the suggestions, however;

* `django_formtools` depends on Django 1.7 or newer (`install_requires=['Django>=1.7'],`), so this breaks compatibility
* Injecting dynamically in install_requires means dropping wheel packaging support. And what to do when no Django is found? Assume we're installing the latest version?

troygrosfield:
@timgraham, with the release of django 1.8, formtools was moved to it's own repo.  How do you guys setup your setup.py to happily work with django 1.8+ as well as older versions of django that require the formtools app? (see discussion above).

timgraham:
Not sure a good solution to this. I'd try asking on the django-users mailing list.

troygrosfield:
@bouke, what about leaving ``django_formtools`` out of the setup.py for now and add doc for django 1.8+ users that a manual install of ``django_formtools`` is needed with a reference to an open ticket (this one?) that explains the situation?  

Then update the ``.travis.yml`` file to conditionally install ``django_formtools`` depending on the django version being tested.  By doing that, the django 1.8 users at least have the ability to install a non-branch version of this repo.

troygrosfield:
I posted to the django-users group.  I'll post the thread here once the topic has been approved.

Bouke:
Thanks everyone for your work. I've included installation instructions in 32eb41d9b0c47f48a9a92f7d5e0ef522c327f28c. In a few months, support for 1.4 is dropped and django-formtools can be included for everyone.

Bouke:
Please let me know if the latest commit on master works for you on Django 1.8. If all is well, I'll release it as 1.2.

troygrosfield:
Just tested the latest master and everything works as expected for me using django 1.8, python 3.4.

Thanks @Bouke! :+1:

PR #106, opened 2015-08-06
Improve validation of phone numbers

jake223:
Use `phonenumbers` to validate phone numbers so that things like '+10' and '003' as well as things like '+123456789100' (too many digits) are not allowed. This used to cause a twilio error, now it returns a nice message. When `phonenumbers` is not installed, revert to the old method of validation.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -10,6 +10,9 @@ django_otp
 qrcode
 # django-formtools on Django 1.8 and above
 
+# Optional - improves phone number validation.
+phonenumbers
+
 # Example app
 
 django-debug-toolbar
--- a/two_factor/migrations/0003_auto_20150806_2156.py
+++ b/two_factor/migrations/0003_auto_20150806_2156.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import two_factor.models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('two_factor', '0002_auto_20150110_0810'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='phonedevice',
+            name='number',
+            field=models.CharField(max_length=16, verbose_name='number', validators=[two_factor.models.phone_number_validator]),
+        ),
+    ]
--- a/two_factor/models.py
+++ b/two_factor/models.py
@@ -9,7 +9,10 @@ from django.utils.translation import ugettext_lazy as _
 from django_otp import Device
 from django_otp.oath import totp
 from django_otp.util import hex_validator, random_hex
-
+try:
+    import phonenumbers
+except ImportError:
+    phonenumbers = None
 try:
     import yubiotp
 except ImportError:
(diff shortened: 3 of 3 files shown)


jake223:
I have to make migrations still for the model changes - I forgot about them because I had only changed validators. I'll make them shortly and then post here - sorry about that.

jake223:
I have written a migration and believe it should work.

dicato:
I like this improvement as it allows us to present more informative error messages to the user. However, allowing `phonenumbers` to be a conditional dependency leads to problems since changing validators on fields requires migrations in Django.

Regardless of whether `phonenumbers` is installed, this migration is going to run. If `phonenumbers` is installed after he migration, you are going to run into problems since the database fields are going to differ from what is described in code.

I believe there are two options to achieve this functionality without the above issue.

* make `phonenumbers` a required dependency
* instead of adding this functionality in the `phone_number_validator` do it somewhere else that does *not* require changing fields or migrations. For example, in `save` on a model or `clean` on a form. I think I'd prefer `clean` on a form if possible.

Bouke:
@jake223 Thanks for the PR. Looks like a solid improvement. @dicato Thanks for the review, those look like valid concerns. 

Regarding the migration; regardless of `phonenumbers` being available, the validator is named `two_factor.models.phone_number_validator`, so the migration shouldn't matter anyway. However this doesn't resolve issue #106, as the validation by `phonenumbers` might not match with Twilio's number validator (of that of any other gateway). Any ideas on how we could resolve those? I'm thinking of putting try/catch around the calling code to the gateway in the view.

dicato:
@Bouke my thought regarding the migration is that the database and code can easily get in an inconsistent state. If you apply this migration without `phonenumbers` installed, then decide to later install it, you will likely have to manually make migrations and migrate.

You are correct this does not fully resolve #109. It should help as it will validate phone numbers when users first enter them, as opposed to when they are used. `try / except` around the calling code makes sense and can likely reuse the same validators to improve the error messages in many scenarios.

It appears that both libraries use the `E.164`(https://en.wikipedia.org/wiki/E.164) standard. See [twilio's error documentation](https://www.twilio.com/docs/errors/21211) and [phonenumbers](https://github.com/daviddrysdale/python-phonenumbers) is a port of Google's [libphonenumber](https://github.com/googlei18n/libphonenumber) - which uses E.164.

Bouke:
I don't think there's an issue with the migrations. Migration 0002 (currently):

    models.CharField(
        max_length=16, 
        verbose_name='number', 
        validators=[django.core.validators.RegexValidator(regex='^(\\+|00)', message='Please enter a valid phone number, including your country code starting with + or 00.', code=b'invalid-phone-number')]
    )

Migration 0003 (new):

    models.CharField(
        max_length=16, 
        verbose_name='number', 
        validators=[two_factor.models.phone_number_validator]
    ),

So the validator is changed, but it's the same `phone_number_validator` regardless of when `phonenumbers` is installed.

However, the tests are failing because `phone_number_validator` is expected to be a RegexValidator, and the message attribute is used in the tests. So when `phonenumbers` is installed, the tests will fail. Furthermore, there are currently no tests in place to validate the added conditional functionality, which should be addressed before I can merge this PR. @jake223 when the tests are adjusted accordingly, I think the PR will be ready for merging.

dicato:
> So the validator is changed, but it's the same phone_number_validator regardless of when phonenumbers is installed.

That causes some unintended side effects. Here is what I just did locally to illustrate the issue:

* `pip install -r requirements_dev.txt` on `master`
* ` cd ./example && python manage.py migrate`
* note that the *new 0003 migration* is applied
* `pip install phonenumbers`
* `python manage.py migrate`
* note the following warning:

```
Your models have changes that are not yet reflected in a migration, and so won't be applied.
Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to
apply them.
```

Installing an *optional* dependency and causing the above warning seems less than ideal. Users of this package are going to end up with additional migrations that aren't part of the package.

Given #109 and #110 I think my previous suggestion of making `phonenumbers` a required dependency is the easiest and most logical path forward.

I'm happy to do this as well as work on the tests. Does this path forward make sense?

Bouke:
You're correct, there would not have been a difference when the validator was defined as a method as well. Anyway, I have written a custom field and validator to resolve this and partially issue #109. You can have a look in the branch [phone-number-validation](https://github.com/Bouke/django-two-factor-auth/compare/features/phone-number-validation). It is not yet complete as there are no unit tests and existing tests are failing, and probably incompatible with Django 1.4. However, could you review it in the current state?

dicato:
:+1: @Bouke I left some comments directly on the commits in that branch!

Bouke:
Closed in favor of PR #112. Can you have a look at my PR? I think it's getting in a good shape now.

PR #107, opened 2015-08-13
Avoid possible staff user / adminsite DoS on sites with many users

pzrq:
...(loading all usernames / user_ids into memory to build the HTML dropdown list) by using raw_id_fields like python-social-auth does.

https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.raw_id_fields
https://github.com/omab/python-social-auth/blob/v0.2.12/social/apps/django_app/default/admin.py#L14

See also django-otp which had a similar admin site issue: 
https://bitbucket.org/psagers/django-otp/pull-requests/14/avoid-possible-staff-user-adminsite-dos-in/diff

### DIFF
--- a/two_factor/admin.py
+++ b/two_factor/admin.py
@@ -76,4 +76,12 @@ if getattr(settings, 'TWO_FACTOR_PATCH_ADMIN', True):
     patch_admin()
 
 
-admin.site.register(PhoneDevice)
+class PhoneDeviceAdmin(admin.ModelAdmin):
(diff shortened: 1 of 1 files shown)


Bouke:
It's a shame that this hasn't been addressed by Django; if a list would contain 1k+ items, default to raw_id_field automatically. Nevertheless, thanks for the PR!

PR #108, opened 2015-08-13
setup.py: pip install_requires breaks with * wildcard blacklist

pzrq:
Possible compromise - blacklist first and last unsupported Django version from each stable/1.5.x and stable/1.6.x branch. 

Feel free to research for other fixes or get it fixed in a new pip, have not looked to see if there is an issue for this blacklist wildcard syntax.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ setup(
     license='MIT',
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
-        'Django>=1.4.2,<1.8.99,!=1.5.*,!=1.6.*',
+        'Django>=1.4.2,<1.8.99,!=1.5,!=1.5.12,!=1.6,!=1.6.11',
         'django_otp>=0.2.0,<0.2.99',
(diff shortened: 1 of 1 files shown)


pzrq:
Also if it only affects local installs feel free to close without taking it.

Bouke:
Can you show me the issue you're having? I believe this is the best solution to communicating that Django 1.5 and 1.6 are not supported. Unless it fails in some way?

Bouke:
Ok I got an issue regarding the version specifiers on pip 1.5.6 which got installed by virtualenv. Upgrading virtualenv to a more recent version installs newer versions of pip/setuptools which can handle these specifiers.

pzrq:
OK, can't reproduce it on a different OSX 10.10 machine which had pip 7.1.0; I have a vague inkling it might have been upgrading pip from 6.1.1 to 7.1.0 or that when I build new virtualenvs on my other machine I still get pip 1.5.6 before upgrading it. 

Closing as invalid, thanks for reviewing.

PR #112, opened 2015-08-18
Phone number validation using python-phonenumbers

Bouke:
This will resolve the following issues:

* Phone number validation will be improved
* Reduced chance of getting error from Twilio
* Phone number formatting
* Local phone number region


two_factor/migrations/0003_auto_20150817_1733.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def migrate_phone_numbers(apps, schema_editor):
+    PhoneDevice = apps.get_model("two_factor", "PhoneDevice")
+    for device in PhoneDevice.objects.all():
+        try:
+            number = phonenumbers.parse(device.number)
+            if not phonenumbers.is_valid_number(number):
+                logger.info("User '%s' has an invalid phone number '%s'." % (device.user.username, device.number))
+            device.number = phonenumbers.format_number(number, phonenumbers.PhoneNumberFormat.E164)
```
dicato: What happens when the phone number is invalid? What does `format_number` return?

dicato:
@Bouke I left some comments throughout the commits. I'm going to do some manual testing today and will report back!

two_factor/migrations/0003_auto_20150817_1733.py
Bouke: Theoretically this shouldn't have happened, as Twilio also validates phone numbers itself. However I wanted to cover the edge-case of logging any invalid phone numbers instead of throwing an exception. I'm not sure how this can be handled more gracefully.

PR #113, opened 2015-08-18
Catch gateway errors

Bouke:
* [x] Show error when gateway raises an exception
* [ ] Disable form input/next when error is shown


Bouke:
Merged in c3c6210a5ee86481ef6db8064afb5d97145c86b4.

PR #121, opened 2015-10-06
Changed to Django's LOGIN_URL

Tusky:
If someone customizes the project and/or has default namespace currently this fails. This does allow customization.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -960,8 +960,8 @@ class UtilsTest(UserMixin, TestCase):
         self.assertEqual(lhs.path, rhs.path)
         self.assertEqual(lhs.fragment, rhs.fragment)
 
-        # We used parse_qs before, but as query parameter order became 
-        # significant with Microsoft Authenticator and possibly other 
+        # We used parse_qs before, but as query parameter order became
+        # significant with Microsoft Authenticator and possibly other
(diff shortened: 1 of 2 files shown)


Tusky:
Did my commit really make FLAKE8 fail? I don't think it could have anything to do with it.

Tusky:
@Bouke any update on if you will merge my branch?

Tusky:
@Bouke could you do any updates on this? it's been a month now without any response :( My admin site still throws error because of this with the original package.

Bouke:
Hi @Tusky I'm not sure why the default shouldn't work for you? If you've customized the package to the extent that the code breaks, you either customized too much or too little. For example, you could include a customized `patch_admin()`, which uses the login view you've defined.

Tusky:
Yes that's what I'm doing now meanwhile (I meant the original project as in if you bump versions that would mess with patch admin or something it might break), however, that is still not a reason why you would not support Django's LOGIN_URL when in other parts you do use it. Hard coding the URL isn't a great solution. In fact the default project set up mentions this:
```
LOGIN_URL = reverse_lazy('two_factor:login')
```
So you are at the exact same place with my code as you are otherwise, except you are allowing further customization when and if needed.

Bouke:
Thanks for your contribution!

Issue #102, opened 2015-06-08
KeyError when trying to login


Bouke:
Seen on the hosted example:

    Internal Server Error: /account/login/
    Traceback (most recent call last):
      File "/django/core/handlers/base.py", line 132, in get_response
        response = wrapped_callback(request, *callback_args, **callback_kwargs)
      File "/django/views/generic/base.py", line 71, in view
        return self.dispatch(request, *args, **kwargs)
      File "/django/utils/decorators.py", line 34, in _wrapper
        return bound_func(*args, **kwargs)
        return view(request, *args, **kwargs)
      File "/django/views/decorators/debug.py", line 76, in sensitive_post_parameters_wrapper
        return func.__get__(self, type(self))(*args2, **kwargs2)
        return bound_func(*args, **kwargs)
      File "/django/views/decorators/cache.py", line 57, in _wrapped_view_func
        response = view_func(request, *args, **kwargs)
      File "/django/utils/decorators.py", line 30, in bound_func
      File "/django/utils/decorators.py", line 34, in _wrapper
        return func.__get__(self, type(self))(*args2, **kwargs2)
      File "/django/utils/decorators.py", line 30, in bound_func
      File "/formtools/wizard/views.py", line 237, in dispatch
        response = super(WizardView, self).dispatch(request, *args, **kwargs)
      File "/django/views/generic/base.py", line 89, in dispatch
        return handler(request, *args, **kwargs)
      File "/two_factor/views/core.py", line 94, in post
      File "/two_factor/views/utils.py", line 123, in post
        return super(LoginView, self).post(*args, **kwargs)
        return super(IdempotentSessionWizardView, self).post(*args, **kwargs)
      File "/formtools/wizard/views.py", line 300, in post
      File "/two_factor/views/utils.py", line 174, in render_done
        return self.render_done(form, **kwargs)
        self.storage.reset()
      File "/formtools/wizard/storage/base.py", line 34, in reset
        wizard_files = self.data[self.step_files_key]
      File "/formtools/wizard/storage/session.py", line 13, in _get_data
        return self.request.session[self.prefix]
      File "/django/contrib/sessions/backends/base.py", line 48, in __getitem__
        return self._session[key]
    KeyError: 'wizard_login_view'

mohsinhijazee:
I am also having the same issue. Is this resolve or any workaround for it?

timfeirg:
I occasionally bump into this issue, refresh page usually fix it

[...]

PR #135, opened 2016-01-08
Check if prefix exists in session before trying to reset storage

vvojvoda:
This pull request addresses the error which occurs when attempting to login a user while another user is already logged in.
This problem is described in issue #102.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -314,6 +314,25 @@ class LoginTest(UserMixin, TestCase):
             "step in the wizard.",
             'token')
 
+    @patch('two_factor.views.utils.logger')
+    def test_login_different_user_on_existing_session(self, mock_logger):
+        """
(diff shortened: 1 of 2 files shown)


Bouke:
Could you also come up with a unit test for this fix? So that future changes will not introduce regressions with respect to this bug. Thanks.

vvojvoda:
Thanks for the review. Added a test.

vvojvoda:
@Bouke Do you think this is good to be merged?

Bouke:
Do you know why the check in [``SessionStorage.__init__``][1] doesn't work?

[1]: https://github.com/django/django-formtools/blob/b11c4375ec3a39c6cd5a5509cb8697936dfc5f50/formtools/wizard/storage/session.py#L8

PR #139, opened 2016-02-04
Fixed migration when a custom usermodel without username field is used

shanx:
Hi,

I have a project with a custom user model that does not have a field called username. This patch changes the migration to dynamically lookup USERNAME_FIELD.

Thanks,
Remco


Bouke:
Thanks for the PR. Instead of inspecting the `USERNAME_FIELD` attribute, shouldn't [`get_username`][1] also work? What would be the benefit of your approach?

[1]: https://docs.djangoproject.com/en/1.9/topics/auth/customizing/#django.contrib.auth.models.AbstractBaseUser.get_username

shanx:
Hey @Bouke you are absolutely right! Thanks for the tip, just updated the PR.

Bouke:
Thank you for your contribution!

PR #142, opened 2016-03-01
Allow to use named urls in LOGIN_URL with django two factor

shanx:
Change https://github.com/Bouke/django-two-factor-auth/commit/3482409f91945c616564f4cf673ce7e4d1f1a799 has django two factor use LOGIN_URL from settings to redirect to. But currently django two factor only allows using literal urls in LOGIN_URL. This can be solved by using reverse_lazy in settings, but since Django itself allows named patterns in LOGIN_URL django two factor should follow. This pulll request fixes this.


Bouke:
Thanks for the contribution, I've merged the PR in d581b0d6ab4c198c61f07cd7a4a463a8f8ba314d.

Issue #149, opened 2016-06-08
Wizard setting to bypass 'welcome' screen


stewbawka:
When onboarding users there's a lot of screens to go through. Would be great to give option to skip initial 'welcome' screen when going through wizard to reduce friction when onboarding users.

moggers87:
Wouldn't it be easier to just pass in your own `form_list` to `SetupView.as_view` or just subclass the view?

PR #150, opened 2016-06-08
TWO_FACTOR_SKIP_WELCOME setting to skip welcome page.

stewbawka:
PR for issue #149 
TWO_FACTOR_SKIP_WELCOME = True in settings will drop welcome page from wizard.
If setting is left out it will continue to work as normal.

Unit tests all passing.

### DIFF
--- a/two_factor/views/core.py
+++ b/two_factor/views/core.py
@@ -204,7 +204,7 @@ class SetupView(IdempotentSessionWizardView):
     template_name = 'two_factor/core/setup.html'
     session_key_name = 'django_two_factor-qr_secret_key'
     initial_dict = {}
-    form_list = (
+    form_list = [
         ('welcome', Form),
         ('method', MethodForm),
         ('generator', TOTPDeviceForm),
@@ -212,7 +212,11 @@ class SetupView(IdempotentSessionWizardView):
         ('call', PhoneNumberForm),
         ('validation', DeviceValidationForm),
         ('yubikey', YubiKeyDeviceForm),
-    )
+    ]
+    if hasattr(settings, 'TWO_FACTOR_SKIP_WELCOME') and settings.TWO_FACTOR_SKIP_WELCOME:
+        form_list.pop(0)
+    form_list = tuple(form_list)
+
     condition_dict = {
         'generator': lambda self: self.get_method() == 'generator',
         'call': lambda self: self.get_method() == 'call',


two_factor/views/core.py
```diff
@@ -204,15 +204,19 @@ class SetupView(IdempotentSessionWizardView):
...
         ('welcome', Form),
         ('method', MethodForm),
         ('generator', TOTPDeviceForm),
         ('sms', PhoneNumberForm),
         ('call', PhoneNumberForm),
         ('validation', DeviceValidationForm),
         ('yubikey', YubiKeyDeviceForm),
-    )
+    ]
+    if hasattr(settings, 'TWO_FACTOR_SKIP_WELCOME') and settings.TWO_FACTOR_SKIP_WELCOME:
```
Bouke: Use `condition_dict` to skip certain wizard pages

two_factor/views/core.py
```diff
@@ -213,7 +213,10 @@ class SetupView(IdempotentSessionWizardView):
         ('validation', DeviceValidationForm),
         ('yubikey', YubiKeyDeviceForm),
     )
+
+    show_welcome = not hasattr(settings, 'TWO_FACTOR_SKIP_WELCOME') or not settings.TWO_FACTOR_SKIP_WELCOME
```
MarkusH: This will be executed at import time, not when the lambda function in the `condition_dict` is being called. Can you either define the lambda expression here or move this into the lambda function's body, please.

MarkusH:
Apart from that, can you please add the settings to the documentation and add a test for this. Thank you :smile:

two_factor/views/core.py
pbassut: why not just not `getattr(settings, 'TWO_FACTOR_SKIP_WELCOME', False)`?

MarkusH: 1. because a definition of `None` in the settings is not the same as `False`.
2. because it will still be executed at import time, not when it should be evaluated.

pbassut: 1. Change `False` to `None` then.
2. What's the problem of being executed at import time?

MarkusH: 2. It's only evaluated once which makes testing harder. When you want to check for different values of `TWO_FACTOR_SKIP_WELCOME` you'd need to take care of that yourself. Using dynamic evaluation at runtime isn't much slower within this simple situation but makes testing easier.

PR #151, opened 2016-07-03
Added docs for monkey patching AdminSite

MarkusH:


Bouke:
The patch should've applied automatically (see https://github.com/Bouke/django-two-factor-auth/blob/8952850329eeb09557702bb087778efaa08b1783/two_factor/admin.py#L53-L66). There should be no need for any additional work. Any ideas why it doesn't work for you by default?

MarkusH:
I don't see how the current implementation enforces 2FA in the admin. You're not overriding the `has_permission()` method.

This doc change is a simple way to keep the existing Django admin site with all its registered models and enforce 2FA there.

Bouke:
Ah yes, now I see what you're trying to do. Seems appropriate!

cpoetter:
I think this does not work anymore. I tried it today with **Django 1.11.15** and I still see admin page shipping with the two factor module, not the standard one from Django.

PR #160, opened 2016-07-13
Fix #159: Redirect to 2FA profile page after removing a phone

### LINKED ISSUE #159: Redirect to profile after deleting a phone
https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/views/core.py#L476

When you delete a phone now, it will take you to `settings.LOGIN_REDIRECT_URL`, which for most people is the homepage.

Would it make more sense to redirect to `reverse('two_factor:profile')`? Like if you want to delete a phone and add a new one? Or delete multiple phones?

MarkusH:


Bouke:
LGTM!

PR #169, opened 2016-08-04
Rework Django admin integration

MarkusH:
- [x] Provide views / URLs under the `/admin/` path
 - [x] login
 - [x] setup
 - [x] backup tokens
 - [x] profile
 - [x] QR code
 - [ ] Phone integration
 - [ ] YubiKey
- [x] Custom Admin styling
- [x] Rework patching of the admin for 2FA


Bouke:
It would be awesome if we could land this PR, as the current admin integration is rather lacking. Do you have plans to continue working on this? If not, I might have a look at this sometime in the next months.

dopry:
@MarkusH, @Bouke do you think this could be largely superseded by https://github.com/jazzband/django-two-factor-auth/pull/491, or is there still value in porting this to the current master?

PR #178, opened 2016-11-08
Upgrade phone number field

emord:
@Bouke 

- Upgrades `phonenumberfield`
- Uses `phonenumberslite` by default (taken from https://github.com/stefanfoulis/django-phonenumber-field/blob/1.1.0/setup.py#L7-L8)
- Removes `debug_toolbar` from `INSTALLED_APPS` as it errors before running the server on django 1.10 due to middleware missing (https://github.com/Bouke/django-two-factor-auth/blob/543687e872bfe7a3d6244f8c73fb9149e53a1df9/example/settings.py#L33)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
         'django_otp>=0.3.4,<0.99',
         'qrcode>=4.0.0,<4.99',
         'phonenumbers>=7.0.9,<7.99',
-        'django-phonenumber-field>=0.7.2,<0.99',
+        'django-phonenumber-field>=1.1.0,<1.99.0',
         'django-formtools',
     ],
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -15,8 +15,8 @@ setup(
         'Django>=1.8',
         'django_otp>=0.3.4,<0.99',
         'qrcode>=4.0.0,<4.99',
-        'phonenumbers>=7.0.9,<7.99',
-        'django-phonenumber-field>=1.1.0,<1.99.0',
+        'phonenumberslite>=7.0.9,<7.99',
+        'django-phonenumber-field>=1.1.0,<1.99',
         'django-formtools',
     ],
     extras_require={
--- a/tox.ini
+++ b/tox.ini
@@ -29,9 +29,9 @@ deps =
     coverage
 
     django-formtools
-    django-phonenumber-field>=0.7.2,<0.99
+    django-phonenumber-field>=1.1.0,<1.99
     django_otp>=0.3.4,<0.99
-    phonenumbers>=7.0.9,<7.99
+    phonenumberslite>=7.0.9,<7.99
     qrcode>=4.0.0,<4.99
     twilio
 ignore_outcome =


example/settings.py
```diff
@@ -65,7 +65,7 @@
     'two_factor',
     'example',
 
-    'debug_toolbar',
+    # 'debug_toolbar',
```
moggers87: Did you try uncommenting the middleware to see if it now works under Django 1.10?

emord: It doesn't. runserver ends with an error `NoReverseMatch: u'djdt' is not a registered namespace`

Seems like making it work should be part of another PR?

Bouke: With version 1.6 of Django Debug Toolbar that should be resolved, can you verify that you're on that version?

Anyway I think its best to separate it from this PR.

setup.py
```diff
@@ -1,4 +1,16 @@
...
+
+install_requires = [
+    'Django>=1.8',
+    'django_otp>=0.3.4,<0.99',
+    'qrcode>=4.0.0,<4.99',
+    'django-phonenumber-field>=1.1.0,<1.99.0',
+    'django-formtools',
+]
+if 'phonenumbers' not in [p[1] for p in pkgutil.iter_modules()]:
+    install_requires.append('phonenumberslite>=7.0.9')
```
Bouke: Can you change this to `...,<7.99`, assuming they semver, would prevent breaking changes.

setup.py
```diff
@@ -1,4 +1,16 @@
 from setuptools import find_packages, setup
+import pkgutil
+
+install_requires = [
+    'Django>=1.8',
+    'django_otp>=0.3.4,<0.99',
+    'qrcode>=4.0.0,<4.99',
+    'django-phonenumber-field>=1.1.0,<1.99.0',
```
Bouke: Can you change this to `...<1.99` instead; so to be consistent with the other version notations?

emord:
@Bouke @MarkusH I looked into the conditional dependencies on wheel, but that didn't seem like the right approach as it wouldn't install either by default which could cause this package to break. Instead I changed the `phonenumbers` dependency to `phonenumberslite`. This way it will only install the necessary dependency if this is the only package that requires `phonenumbers`, but a user can also install the full package on top of it. Thoughts?

MarkusH:
I'm trying to figure out the difference between `phonenumbers` and `phonenumberslite`. The latter seems to have the same size when you look at the `.whl` file. The `.tar.gz` source file is considerably smaller, though. But since pip prefers wheel packages these days I don't see a benefit in actually changing those dependencies ... . Can you give some background why you would like to have this changed?

emord:
@MarkusH https://github.com/daviddrysdale/python-phonenumbers#memory-usage

The last paragraph in that section describes the difference between packages. I'm not too familiar with `wheel` vs `tar.gz` packages to be honest. Before `7.6.0` there was only the `tar.gz` format which is supported here.

I proposed this change as `phonenumberslite` as we use it in our repository and ran into an issue with our deploy process when this installed `phonenumbers` (which we fixed). There wasn't a need for the full package and the field package was behind as well, so I figured it wouldn't hurt

Bouke:
I've updated the dependency. As the default dependency library for django-phonenumber-field changed to phonenumberslite, there is no longer a need to replicate the dependency in this package. So I've taken it out.

emord:
@Bouke Thanks! Could you release a version with that update?

For what it's worth, I think it may have been dangerous (although unlikely to cause issues) to take out `phonenumbers` as a dependency. It's used explicitly here https://github.com/Bouke/django-two-factor-auth/blob/f308883d7e1e327f06080da254c75a20e5fcc4fb/two_factor/migrations/0003_auto_20150817_1733.py#L6 and some other places. I think if that is removed from a dependency in `django-phonenumber-field` this library could cause error

PR #180, opened 2016-11-25
Add support for image URLs

mmazur:
I know at least FreeOTP for android supports this.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -71,6 +71,11 @@ General Settings
   audience is a certain country, setting the region to that country allows
   entering phone numbers without that country's country code.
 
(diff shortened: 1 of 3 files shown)


moggers87:
Informal review: could do with a few simple unit tests.

Bouke:
This appears to be a non-standard extension of the [Key Uri Format](https://github.com/google/google-authenticator/wiki/Key-Uri-Format). Is this extension in common use on larger websites? It might break other OTP applications that don't expect that URL parameter.

mmazur:
Hm, I sort of assumed this is what 2fa apps that allow selection of icons supported, but seems not and this thing is only present in android's version of freeotp. I'm guessing this of itself is enough to not get the patch merged?

PR #183, opened 2016-12-22
Fix #182: Autofocus token input during authentication

### LINKED ISSUE #182: Autofocus token input during authentication
Opening as an issue first for any background discussion before a PR.

Token input usually comes directly after user:pass input.  The user already has their hands on the keyboard and the token is the only field for input in the wizard flow.  It should auto focus, so no additional tab or mouse clicks are necessary to complete authentication.

While this could be a template decision, since the authentication workflow is wrapped in a wizard, I think changing the AuthenticationTokenForm makes the most sense.

```
grind@hollywood:~/git/django-two-factor-auth/two_factor$ git diff
diff --git a/two_factor/forms.py b/two_factor/forms.py
index 7e61ce9..57dbc90 100644
--- a/two_factor/forms.py
+++ b/two_factor/forms.py
@@ -142,6 +142,8 @@ class AuthenticationTokenForm(OTPAuthenticationFormMixin, Form):
     otp_token = forms.IntegerField(label=_("Token"), min_value=1,
                                    max_value=int('9' * totp_digits()))
 
+    otp_token.widget.attrs.update({'autofocus': 'autofocus'})
+
     # Our authentication form has an additional submit button to go to the
     # backup token form. When the `required` attribute is set on an input
     # field, that button cannot be used on browsers that implement html5
```

sdann:
Token input usually comes directly after user:pass input. The user
already has their hands on the keyboard and the token is the only field
for input in the wizard flow. It should auto focus, so no additional tab
or mouse clicks are necessary to complete authentication.

While this could be a template decision, since the authentication
workflow is wrapped in a wizard, I think changing the
AuthenticationTokenForm makes the most sense.

### DIFF
--- a/two_factor/forms.py
+++ b/two_factor/forms.py
@@ -142,6 +142,8 @@ class AuthenticationTokenForm(OTPAuthenticationFormMixin, Form):
     otp_token = forms.IntegerField(label=_("Token"), min_value=1,
                                    max_value=int('9' * totp_digits()))
 
+    otp_token.widget.attrs.update({'autofocus': 'autofocus'})
+
     # Our authentication form has an additional submit button to go to the
     # backup token form. When the `required` attribute is set on an input
     # field, that button cannot be used on browsers that implement html5


sdann:
Anything else needed on the PR before acceptance?

Bouke:
I needed a second look to where that Travis failure was coming from. But it's not related to your change. I'll resolve that with another commit. Thanks for your contribution.

sdann:
Sounds good.  Thank you.

PR #187, opened 2017-01-07
Change DisableView to match release notes

moggers87:
Looks like this was missed when changing redirect_url to success_url in
other views


Bouke:
Looks like an oversight indeed, thank you!

PR #192, opened 2017-02-09
Fix failure if LOGIN_REDIRECT_URL is an URL

PierreF:
If LOGIN_REDIRECT_URL is an URL (and not a named URL), adding or removing an backup phone will generate an error: ```Reverse for '/accounts/login/' with arguments '()' and keyword arguments '{}' not found. 0 pattern(s) tried: []```

This is due to PhoneSetupView and PhoneDeleteView that pass LOGIN_REDIRECT_URL to reverse_lazy. But reverse_lazy only works with an named URL or a view function.

To reproduce this issue, changing tests.settings.LOGIN_REDIRECT_URL from "two_factor:profile" to "/account/two_factor/" will reproduce this issue when running tests.

### DIFF
--- a/tests/test_views_phone.py
+++ b/tests/test_views_phone.py
@@ -17,6 +17,7 @@ from django_otp.util import random_hex
 from two_factor.models import PhoneDevice
 from two_factor.utils import backup_phones
 from two_factor.validators import validate_international_phonenumber
+from two_factor.views.core import PhoneDeleteView, PhoneSetupView
 
 from .utils import UserMixin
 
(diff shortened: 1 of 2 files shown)


moggers87:
@PierreF tests?

PierreF:
To add test, I have to change LOGIN_REDIRECT_URL and it must be done very early because when PhoneSetupView is defined, it will capture the current value of settings.LOGIN_REDIRECT_URL).

Changing later (using @override_settings or similar) has no effect.

The only clean solution I think of is to add another test flavour for tox and use environment to change the value of LOGIN_REDIRECT_URL. In a similar way custom_user is done. But this will add more tests combination... does it worth it ?

Another idea could be to monkey-patch two_factor.views.core.PhoneSetupView.success_url in a test, something like:
```
with self.settings(LOGIN_REDIRECT_URL='/account/two_factor/'):
    two_factor.views.core.PhoneSetupView.success_url = settings.LOGIN_REDIRECT_URL
    # Test
```
But it's not very clean... the test-case monkey-patch the code with a (partial) fix.

moggers87:
Surely just setting values (both by plain urls and reverse_lazy) of success_url and then calling `get_context_data`/`get_success_url` to check the output would be enough? `LOGIN_REDIRECT_URL` is only the default value and doesn't need to be touched (in my opinion)

PierreF:
I've added some tests. Are you thinking to tests like those ?

get_success_url only exists with the DeleteView not with the WizardView. So for PhoneSetupView I only test get_context_data.

moggers87:
You should also be testing with `LOGIN_REDIRECT_URL = reverse_lazy('blah')` (because people will probably want that) and with URL names `LOGIN_REDIRECT_URL = 'blah'` (for backwards compatibility testing)

Sorry for the late reply.

PierreF:
It is good with latest changes or something is still missing on the tests ?

moggers87:
Had a quick look at your latest commit, looks good.

Bouke:
Thank you for your contribution!

PR #193, opened 2017-02-21
Rewritten login view without IdempotentSessionWizardView inheritance …

koleror:
In some condition (probably when the log in is aborted), the login and password may be saved in the session as plain text and are never cleared. This was due to the use of the SessionWizardView which saves the full form in the session to remember it's state for the next step. 

To fix that, I rewritted the LoginView without using the WizardView. To make it work, I forward the entered  login and password in the token form as hidden fields (AWS does pretty much the same using ajax).

I also overrided the `get_template_names` method to be able to use different templates depending on the step. By default, I therefore changed the auth template to the default Django one but this can be overrided.

I tested my changes with Google Authenticator only, it may not work properly on other devices. Help wanted to test that :) 
The backup tokens also works correctly. 
Any feedback would be appreciated!


two_factor/views/core.py
```diff
@@ -159,29 +175,36 @@ def get_user(self):
...
-
-        context['cancel_url'] = resolve_url(settings.LOGOUT_URL)
+        context = super(LoginView, self).get_context_data(**kwargs)
+        self.step_cache = step
+        context['step'] = step
+        if step == 'auth':
+            context['form'] = form or self.form_dict['auth']()
+        elif step in ['token', 'backup']:
+            context['username'] = self._username_cache
+            context['password'] = self._password_cache
```
moggers87: I think this just moves the issue of storing a password in the clear rather than solving it.

koleror: Yes, but in the instance only, instead of the database.

moggers87: What about someone logging in on a public computer? What if the browser mistakenly caches the page?

koleror: I don't think browsers do cache forms, especially when `autocomplete=False`

setup.py
```diff
@@ -12,7 +12,7 @@
     license='MIT',
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
-        'Django>=1.8',
+        'Django>=1.7',
```
MarkusH: Why did you change this to 1.7? django-two-factor-auth doesn't support 1.7. Django 1.7 doesn't get any security updates anymore either.

setup.py
```diff
@@ -28,6 +28,7 @@
         'Development Status :: 5 - Production/Stable',
         'Environment :: Web Environment',
         'Framework :: Django',
+        'Framework :: Django :: 1.7',
```
MarkusH: Remove this as well, please.

two_factor/forms.py
```diff
@@ -138,7 +139,7 @@ class DisableForm(forms.Form):
     understand = forms.BooleanField(label=_("Yes, I am sure"))
 
 
-class AuthenticationTokenForm(OTPAuthenticationFormMixin, Form):
+class AuthenticationTokenForm(forms.Form, OTPAuthenticationFormMixin):
```
MarkusH: Mixins always go before forms, please.

two_factor/forms.py
```diff
@@ -165,11 +166,13 @@ def __init__(self, user, initial_device, **kwargs):
         # YubiKey generates a OTP of 44 characters (not digits). So if the
         # user's primary device is a YubiKey, replace the otp_token
         # IntegerField with a CharField.
-        if RemoteYubikeyDevice and YubikeyDevice and \
-                isinstance(initial_device, (RemoteYubikeyDevice, YubikeyDevice)):
+        YubikeyDevices = (RemoteYubikeyDevice, YubikeyDevice)
```
MarkusH: This is an unrelated change to your reported security issue as well.

two_factor/views/core.py
MarkusH: _Never_ underestimate what browsers might do or not, please. I don't understand why we need to store the password in some context at all? After a successful authentication with username and password a user is authenticated. The verification via 2FA is a secondary step that happens _after_ a successful authentication.

moggers87: @MarkusH user authentication actually happens more than once - any time `get_user` is called on the login view the username and password are fetched from the session and then passed to the authentication form. The user is authenticated one final time during `done` when `login` is called: https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/views/core.py#L106

We shouldn't really be storing the user's password _anywhere_ IMO. There are, for example, some people who like to use the signed cookie session backend, which again leaves passwords in the clear and on the browser.

koleror: @moggers87 You're right. The auth is done at every step of the workflow. That's why we need to save the username/password in some way, even if I'm not a big fan of it either. With my fix, at least I don't save it in the DB anymore, it's only sent back to the user as hidden form input, to make the user send it again with the token form seamlessly. If you have a better idea, tell me :)

koleror:
@MarkusH I repushed to revert anything that is not directly related to the security fix. I''ll do another PR later for it later.

two_factor/views/core.py
moggers87: @koleror how about not saving the username and password at all? If we've verified it once, why do we need to verify it again?

We could have something like this:

1. do the first step as normal
2. store the user id
3. do the second factor as normal
4. fetch the user via their id and call `login`

Maybe we should open an issue to properly discuss alternatives?

koleror: Hmm I thought about that but thought that if nobody do that, there must be a reason. 
AWS also sends back the credentials to the user to make him send them back with the token on the second form. So I tend to think they must be right. I'm trying to save the password back to the session, but in an encrypted way this time. I'll let you know how it goes.

MarkusH: I think sending the password back and forth doesn't fix the issue in any proper way. There should be a way to only do the authentication _once_ and not on every step.

MarkusH: > I'm trying to save the password back to the session, but in an encrypted way this time. I'll let you know how it goes.

No, please don't. Set a flag `user_was_successfully_authenticated_by_password` -- or whatever you want to name -- it in the session and check for that.

> AWS also sends back the credentials to the user to make him send them back with the token on the second form (the only difference is they do it using ajax).

No, AWS does not send back the password. They send the initial authentication request via XHR, store a successful authentication in the session and then send another POST request with the MTA token which happens to include the password again.

moggers87: > AWS also sends back the credentials to the user to make him send them back with the token on the second form (the only difference is they do it using ajax).

If they're using ajax, the password is probably kept client side for the second form and then inserted via JS (I really can't imagine any reason why you'd want to send a user back their own password). That would actually be a good way of doing things, but would require a major rewrite of the login view.

koleror: ok, I made an alternate implementation following your suggestion here: https://github.com/Bouke/django-two-factor-auth/pull/194

What do you think?

Bouke: The reason why the username/password are currently stored in the session, is to enforce that the login only happens after both factors have been verified.

moggers87:

PR #194, opened 2017-02-22
[Implementation B] Rewritten login view without IdempotentSessionWizardView inheritance

koleror:
After discussion there : https://github.com/Bouke/django-two-factor-auth/pull/193, here is my second proposal which saves in the session the fact that the user passed the auth step to bypass it on the token step.


koleror:
@MarkusH @moggers87 Any feedback?

two_factor/forms.py
```diff
@@ -176,3 +177,31 @@ def clean(self):
 
 class BackupTokenForm(AuthenticationTokenForm):
     otp_token = forms.CharField(label=_("Token"))
+
+
+class TokenAuthenticationForm(AuthenticationForm):
```
koleror: Because I wanted to keep using the normal login form behaviour (especially the `confirm_login_allowed` step + I wanted to make it transparant for the view whether the user is properly authenticated or it is a fake login

two_factor/forms.py
```diff
@@ -176,3 +177,31 @@ def clean(self):
...
+    backend = forms.CharField()
+
+    def __init__(self, *args, **kwargs):
+        super(TokenAuthenticationForm, self).__init__(*args, **kwargs)
+        del self.fields['username']
+        del self.fields['password']
+
+    def authenticate(self):
+        UserModel = get_user_model()
+        user = UserModel.objects.get(id=self.cleaned_data['user_id'])
```
koleror: Hmm that's a good point, but you asked me to *only* add the security fix related code in this PR, so I guess it could be an improvement for later as this was already the case before. 
+ The hacker would also have to access the user's computer which would still be on the token form page as I don't think there is any way to access the token page directly. 
Also, accessing the auth page clears the session info.
What do you think?

two_factor/views/core.py
```diff
@@ -48,7 +49,7 @@
 
 @class_view_decorator(sensitive_post_parameters())
 @class_view_decorator(never_cache)
-class LoginView(IdempotentSessionWizardView):
+class LoginView(TemplateView):
```
koleror: Because I wanted to make sure the security issue caused by the inheritance was fixed and would not come back by accident.

koleror:
@moggers87 I have indeed seen that the tests were not passing but I waited for an approval before cleaning things up.  
> "I don't feel good about the implementation in general.". Why ? Can you develop?

moggers87:

PR #204, opened 2017-04-10
Fixed redirect_to after successful login

FabioFleitas:
The `redirect_to` was not working correctly on a successful POST request login. I updated the logic to be consistent with how Django does it. See [here](https://github.com/django/django/blob/master/django/contrib/auth/views.py#L69) as example to how Django handles getting `redirect_to` from params.

### DIFF
--- a/tests/test_views_login.py
+++ b/tests/test_views_login.py
@@ -53,6 +53,15 @@ class LoginTest(UserMixin, TestCase):
              'login_view-current_step': 'auth'})
         self.assertRedirects(response, redirect_url)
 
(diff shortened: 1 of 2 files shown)


FabioFleitas:
@Bouke I went ahead and added a `test_valid_login_with_custom_post_redirect` test case. Let me know if there's any other changes you'd like.

Bouke:
Nice work, thank you!

FabioFleitas:
No problem! When will you deploy this to PyPI?

Bouke:
I have not date for it; as a deploy takes quite a bit of my free time.

PR #206, opened 2017-04-17
Fixed the twillo gateway to work with the newest version of the twillio client

ribalba:


Bouke:
It took me some time to realize why the tests failed; `pip_pre` was set to `True`, which installed twilio 6.3.dev0 (from 2016), while 6.0.0 (from 2017) should've been installed. It seems that 6.3.dev0 was uploaded to pypi by accident, as it's older and the old API.

I've merged your PR, thank you.

PR #212, opened 2017-05-31
Fix Twilio v6 usage

maroux:
- Twilio v6 library is backwards incompatible
- Machine detection has been changed to work based on Webhook, and rather than figure out appropriate response to hang up, I chose to disable machine detection.

Fixes Bouke/django-two-factor-auth#211

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,8 +19,8 @@ setup(
         'django-formtools',
     ],
     extras_require={
-        'Call': ['twilio'],
-        'SMS': ['twilio'],
+        'Call': ['twilio>6.0'],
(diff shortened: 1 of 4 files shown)


Bouke:
Thank you for your contribution, merged in 94866ef786eba79b8b9bc7997e38dbdcafef3ff9.

At first I thought we needed to re-enable Answer Machine Detection; however:

> The public beta for Answering Machine Detection supports calling to the US, Canada, United Kingdom, Netherlands, Finland and Sweden.

They removed the `IfMachine` feature, replacing it with a limited and priced ($0.0075 per call) beta.

PR #215, opened 2017-06-11
Add U2F devices feature

### LINKED ISSUE #86: integrate django-u2f
I would be great to integrate: https://github.com/gavinwahl/django-u2f

What would it take?

vereena42:
I wanted to resolve #86 issue and integrate django-u2f with this project, but after investigation I realized these two projects differ too much to be integrated.
So instead I wrote code for using U2F devices in this project from scratch.
Now one can choose U2F in setup view and login with U2F. There is also new functionality in profile view - if U2F is default method one can add more than one U2F keys to their account and delete keys - after the last one is deleted two_factor auth is disabled.

Plese don't hestitate to review my code - I'm open for any advice.


Bouke:
Hi @vereena42 sorry for not getting back to you any sooner, but nevertheless great seeing you as a first time contributor!

I'm not familiar with U2F, but I've ordered a Yubikey some years ago. I must still have it somewhere. The idea of Yubikey was nice, however I haven't found any use-case for it myself as most services I'm using don't support it. U2F seems like a device similar to Yubikey, and based on my experiences thus far I'm a bit sceptical. However support for U2F has come up a few times, so there might be others that are interested.

So going forward I think integration of U2F should be similar to Yubikey: completely optional. Further I think we should not introduce additional views, but re-use existing views. Is there a specific reason you introduced new views? And most importantly we need unit tests before any code can be merged in.

jeruyyap:
To clarify the use-case for U2F, the biggest advantage of U2F over TOTP apps and Yubico OTP is that it provides a degree of protection against phishing and makes man-in-the-middle considerably more difficult.

A lot of sites and applications that use U2F allow you to use as an option alongside a TOTP app because U2F support on mobile is limited, but this is still makes sense because logins via U2F are protected against phishing and are much more difficult to man-in-the-middle compared to logins using the TOTP app, so using the U2F whenever possible even if TOTP app fallback is still needed occasionally still reduces the user's overall chances of being phished or MITM-ed. 

It's notable that Yubico acknowledges the advantages of U2F over Yubico OTP:
https://www.yubico.com/2016/02/otp-vs-u2f-strong-to-stronger/ 
And now sell a device that only supports U2F:
https://www.yubico.com/product/fido-u2f-security-key/

alexsdutton:
I'm very much interested in this too. I'd just started writing a django-otp device plugin for U2F (https://github.com/alexsdutton/django-otp-u2f), and was looking to integrate it with django-two-factor-auth when I found this.

I'd be very happy to help in making this happen.

My thoughts:

* Should the device implementation live outside of django-two-factor-auth?
* Could integrating new devices into django-two-factor-auth be made pluggable? (at the moment `SetupView.get_method()` is effectively hard-coded with device types, making adding new ones from outside a bit awkward)

Bouke:
@alexsdutton  Regarding your questions;

1. the implementation could live within this package, however in a similar fashion to yubikey i.e. completely optional to use
2. I'd very much like to move this package to a pluggable architecture. That way we could add support for additional devices, like u2f and others. Even the current built-in phone device (sms / call) could move to a plugin. This package's core would then only implement the plumbing to connect those packages and plugins.

Braintelligence:
It would be very awesome if U2F would be supported. Due to the nature of the device key never leaving the U2F-device and the rising availability of NFC-capable U2F-devices, U2F-keys are getting more and more popular as a means of 2FA.

Please consider integrating this soon 😸.

PR #216, opened 2017-07-01
Update setup.py

Temeez:
Fixes: `ValueError: ("Expected ',' or end-of-list in", 'Django >=1.8,!=1.9.*', 'at', '*')` when trying to install via pip.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ setup(
     license='MIT',
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
-        'Django>=1.8,!=1.9.*',
+        'Django>=1.8,!=1.9',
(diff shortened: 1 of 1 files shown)


Bouke:
Hi @Temeez, thanks for reaching out. What issue are you seeing, under which conditions (versions of Python / pypi etc.). This hasn't been an issue for me, and it's the first time it's brought to my attention. So I'm curious why it only fails now for you, as the code has been there for about 3 months now. I'm not sure about the correct syntax here, so I don't want to break it for somebody else with this PR.

Temeez:
My bad, pip was 1.5.6 (oldish vagrant dev env). There seems to be no problem with the syntax after upgrading pip to the most recent one.

PR #218, opened 2017-07-20
Make sms body configurable

snoepkast:
Because we would like to set a custom body for the sms message.


Bouke:
I was under the impression that using a custom i18n file (.po/.mo) you would be able to specify a custom translation in any language of your choosing. Including English. Would that work for you?

snoepkast:
We don't use translations, so setting that up just for a simple text change might be a bit overkill I think.

Bouke:
While I can see your reasoning, I'm reluctant to add a configuration for this. We already have a way to customise the text, so adding another for a specific use-case seems too much. If you really want to do this, you could copy the twilio gateway code, customise and point to from your configuration.

PR #232, opened 2017-10-30
Added SMS message template in django settings for twilio

janadh:
Added reading text message body template from the django settings file (if specified). If not specified then the default value will be used

## Description

Enables the users to use django settings TWILIO_SMS_BODY_TEXT to have their own custom text message text. 

## Motivation and Context

At present the users cannot have their own custom text message while sending authentication messages

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:


Bouke:
See #218 which proposes something similar and the reasoning on why it wasn't merged.

PR #240, opened 2018-01-04
Use reverse_lazy

danielquinn:
Fixes a reference in `settings.py` to call `reverse_lazy()`

## Description

I'm thinking this was an oversight, right?  Or would this somehow work without actually calling `reverse_lazy()`?

## Motivation and Context

Reading through the documentation, this didn't make sense to me.

## How Has This Been Tested?

It's not.

## Types of changes

## Checklist:

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -39,8 +39,6 @@ Point to the new login pages in your ``settings.py``:
 
 .. code-block:: python
 
-    from django.core.urlresolvers import reverse_lazy
-
     LOGIN_URL = 'two_factor:login'
 
     # this one is optional


moggers87:
`django.shortcuts.resolve_url` is used in all cases where I could find those settings being used. It can handle both proper URLs and URL names.

Maybe a comment could be added to the docs saying that you can either use `reverse_lazy` if you prefer?

danielquinn:
Huh.  You learn something new every day :-)

I guess the only thing that's confusing is that you import `reverse_lazy` there and then don't use it.  So I'd suggest either dropping the import or adding the comment as you suggest.  If you pick one, I can amend the PR if you like.

moggers87:
I'd go for removing the import.

danielquinn:
Done :-)

Bouke:
The import was a left-over after prior changes, nice find, thank you!

PR #247, opened 2018-02-06
Replace ValidationError with SuspiciousOperation in views

### LINKED ISSUE #243: Post requests with missing management form data to LoginView cause 500 errors
If a client submits a username and password, but for some reason *doesn't* also send "login_view-current_step" (let's say it's a bot sweeping the Internet for login forms and then trying common username/password combinations), an exception is raised.

## Expected Behavior

Either a client error (HTTP 400?) or a form error as would happen with an invalid username/password/token.

## Current Behavior

A `ValidationError` is raised but not caught. The causes a 500 error and an alert for developers/admins.

## Possible Solution

Replace `ValidationError` with `SuspiciousOperation`. This exception gets logged with `django.security.*` rather than `django.request`, so it can be processed differently if admins prefer: https://docs.djangoproject.com/en/2.0/topics/logging/#django-security

1. Go to your login page
2. Use the HTML inspector to remove the "login_view-current_step"  hidden input
3. Submit the form and see you have received a 500 error response

## Context

The current error produces an error that looks like a server error rather than a client error. I think this bug is also present in django-formtools, but both LoginView and IdempotentSessionWizardView override the post method and manually raise a `ValidationError`.

## Your Environment

* Browser and version: Firefox 57
* Python version: Python 2.7.14
* Django version: Django 1.11.9
* django-otp version: 0.4.2
* django-two-factor-auth version: 1.7.0
* Link to your project: https://github.com/Inboxen/Inboxen

moggers87:
## Description

Nothing catches the ValidationError, causing a 500 error to be given. SuspiciousOperation is caught by Django and a HTTP 400 response is returned to the client.

## Motivation and Context

Fixes #243 

## How Has This Been Tested?

Tests added for the various places where the Wizard view is used. Build passed on travis: https://travis-ci.org/moggers87/django-two-factor-auth/builds/338217926

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_views_login.py
+++ b/tests/test_views_login.py
@@ -262,6 +262,14 @@ class LoginTest(UserMixin, TestCase):
                                'login_view-current_step': 'auth'})
         self.assertRedirects(response, resolve_url(settings.LOGIN_REDIRECT_URL))
 
+    def test_missing_management_data(self):
+        # missing management data
+        response = self._post({'auth-username': 'bouke@example.com',
+                               'auth-password': 'secret'})
+
+        # view should return HTTP 400 Bad Request
+        self.assertEqual(response.status_code, 400)
+
 
 class BackupTokensTest(UserMixin, TestCase):
     def setUp(self):
--- a/tests/test_views_phone.py
+++ b/tests/test_views_phone.py
@@ -124,6 +124,14 @@ class PhoneSetupTest(UserMixin, TestCase):
         self.assertIn('cancel_url', context)
         self.assertEqual(url, context['cancel_url'])
 
+    def test_missing_management_data(self):
+        # missing management data
+        response = self._post({'setup-number': '123',
+                               'setup-method': 'call'})
+
+        # view should return HTTP 400 Bad Request
+        self.assertEqual(response.status_code, 400)
+
 
 class PhoneDeleteTest(UserMixin, TestCase):
     def setUp(self):
--- a/tests/test_views_setup.py
+++ b/tests/test_views_setup.py
@@ -228,3 +228,10 @@ class SetupTest(UserMixin, TestCase):
         with self.settings(TWO_FACTOR_SMS_GATEWAY='two_factor.gateways.fake.Fake'):
             response = self.client.get(reverse('two_factor:setup_complete'))
             self.assertContains(response, 'Add Phone Number')
+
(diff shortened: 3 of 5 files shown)


moggers87:
It might be worth waiting to see what upstream has to say about the same issue in django-formtools: https://github.com/django/django-formtools/pull/119

Bouke:
The current implementation is a [copy-paste from upstream](https://github.com/django/django-formtools/blob/164f0f46e651292680204f3c2a1afbaf9b31085c/formtools/wizard/views.py#L286-L289), with certain modifications to make the form wizard idempotent. Ideally I'd like to stay as close to the upstream code as possible to keep maintaining our patches low.

In #243 you also specified the problems the behaviour of throwing a `ValidationError` is causing you. You said that if the field is missing, due to the throwing of the `ValidationError`, this results in a 500 error. Is this not the case when a `SuspiciousOperation` is thrown?

Is there a way for this package to catch the ValidationError and show some error message instead? Can we put a try-except somewhere?

moggers87:
An alternative approach would be to not raise a validation error at all and do `return self.render(form)`. Something (the management form?) would have to added to the template context as well to display an error message.

I don't like that idea for two reasons:

* Nothing will get logged (currently it's an error to `django.request`, this PR would change that to it being logged to `django.security`)
* It changes what the user sees from "something went wrong and I can't fix it" to "something went wrong and it's saying I need to fix it but I don't see how"
  * Assuming the developer made a mistake in their custom template rather than actual user tampering

My goal with this PR was only to change what the site admin sees (a code error vs. a client not sending all the required data for some reason) and not to change what the user sees.

moggers87:
> In #243 you also specified the problems the behaviour of throwing a ValidationError is causing you. You said that if the field is missing, due to the throwing of the ValidationError, this results in a 500 error. Is this not the case when a SuspiciousOperation is thrown?

SuspiciousOperation results in a 400 response. not a 500. IIRC this still sends an email to admins, but it points more clearly to an error from data being sent from the client rather than the view code.

moggers87:
This has been merged upstream: https://github.com/django/django-formtools/pull/119

I've also rebased this branch on latest master.

Bouke:
I think it would be good to mention this in the change log. Can you add a mention of this change?

Bouke:
And thanks for the PR!

Issue #221, opened 2017-08-23
Authentication backend's authenticate() method is called 3 times during a successful authentication


hsenot:
## Current Behavior

I am using this package with a custom authentication backend and logging the calls to the authenticate() method. It seems there are 3 calls to authenticate() during a successful authentication (instead of 1).

## Possible Cause

The formtools wizard post() method called from IdempotentSessionWizardView's [post()](https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/views/utils.py#L138) contains 3 direct or indirect calls to form.is_valid():

1. `if form.is_valid():` [here](https://github.com/django/django-formtools/blob/master/formtools/wizard/views.py#L301)

2. `self.steps.last` [here](https://github.com/django/django-formtools/blob/master/formtools/wizard/views.py#L307)

3. `return self.render_done(form, **kwargs)` [here](https://github.com/django/django-formtools/blob/master/formtools/wizard/views.py#L309)

Each of these 3 calls is triggering a form.is_valid() call, and as a consequence, an authentication backend authenticate() call.

I am a bit unsure if this issue is better addressed in this project, in formtools or in contrib.auth's AuthenticationForm. I think the root cause is the use of contrib.auth's AuthenticationForm within the formtools wizard.

## Steps to Reproduce (for bugs)

1. Create a custom authentication backend based on the default ModelBackend https://github.com/django/django/blob/master/django/contrib/auth/backends.py#L7-L103
2. Log calls to authenticate()
3. Login successfully (even with a non-2FA user) and check the log for 3 calls to authenticate()

## Context

In the context of a remote user authentication backend, 3 external API calls to a remote server can add significantly to the time needed for logging in - it degrades performance and scalability.

## Your Environment

* Python version: 3.4
* Django version: 1.11.1
* django-two-factor-auth version: 1.6.2

Bouke:
There have been talks about replacing the LoginView with a custom design, dropping the Wizard from formtools. For normal use-cases the additional calls to authenticate should not be a problem, but for certain use-cases (like yours) I can see that the current situation is not ideal. However replacing the form isn't something that is actively being worked on (as far as I know), but two proposals have been made: #193 and #194.

moggers87:
I believe this has been fixed in 1.12.0.

PR #249, opened 2018-02-19
Store the user's password hash in the session rather than plain text

### LINKED ISSUE #239: Passwords are stored in plain-text in the user's session
When using the login wizard the user's password is stored in plain text in the session store. The information is stored by formtools' wizard class. This is a security risk, depending on the deployment this can easily lead to a leak. Imagine a shared database ...

The information is not required to complete the token step, it would suffice to store the authentication state once the user's username and password are verified and then call `login()` in the `done()` method.

thecardcheat:
## Description

This maintains the current behavior of allowing the wizard to re-validate the auth form but keeps the user's password out of the session.

## Motivation and Context

This is a more straightforward solution than either #193 or #194 and should close #239 and #221. It's possible that certain configurations may not want to allow the password hash into the session either, but in many cases the level of access to the table with password hashes will match those with access to the session and it is still much more secure than allowing the plain text password to persist at all.

## How Has This Been Tested?

Ran tox suite and tested in a local project.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/forms.py
+++ b/two_factor/forms.py
@@ -2,6 +2,10 @@ from binascii import unhexlify
 from time import time
 
 from django import forms
+from django.contrib.auth import get_user_model
+from django.contrib.auth.forms import (
+    AuthenticationForm as DjangoAuthenticationForm,
+)
 from django.forms import Form, ModelForm
 from django.utils.translation import ugettext_lazy as _
 from django_otp.forms import OTPAuthenticationFormMixin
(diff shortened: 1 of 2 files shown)


Bouke:
This assumes a `password` field on the user model, which I think is not a mandatory field on a [`CustomUser`](https://docs.djangoproject.com/en/2.0/topics/auth/customizing/#django.contrib.auth.models.CustomUser). An application might be backed by LDAP for instance, so the password is stored external to this application. 

Also I'm a bit wary as this adds a custom authentication form and custom authentication logic, which might not have received the same scrutiny as Django's built-in login form. (I'm aware that the same argument also holds for this library itself.)

thecardcheat:
I figured non-PW back-ends would need to provide their own Form anyway since `django.contrib.auth.forms.AuthenticationForm` is just a password auth form, but I can see with this course better checking of the form type would need to be properly handled in `get_form_step_data()`.

I think the form is fine as-is since you'd have to provide one that matches your auth otherwise, but I have never used a non-PW auth back-end in Django.

> I'm a bit wary as this adds a custom authentication form and custom authentication logic, which might not have received the same scrutiny as Django's built-in login form. (I'm aware that the same argument also holds for this library itself.)

That's fair but I think it is worth the maintainability burden because saving the password in plain-text is a deal-breaker for a lot of projects, since it transiently defeats the purpose of using a strong hash, and does so when using an app you'd expect to be increasing security. Implementation-wise it's straightforward and follows the existing `AuthenticationForm` form as closely as possible.

Bouke:
> I figured non-PW back-ends would need to provide their own Form anyway since `django.contrib.auth.forms.AuthenticationForm` is just a password auth form, but I can see with this course better checking of the form type would need to be properly handled in get_form_step_data().

I suspect that custom authentication backends can just piggy-back on the default authentication form. It isn't directly tied to the database layer, but calls `authenticate` instead:

https://github.com/django/django/blob/fe99fb860f12c7440752441ef0da37ed65e5bbe3/django/contrib/auth/forms.py#L192

The issue here is not that other authentication backends might not be expecting a password, but the current implementation that (a) expects the user accounts to be stored in the local database and (b) expects a field called "password" that contains the password hash in a certain encoding. This will cause problems for users of this library that have a custom password validation backend.

anlutro:
Maybe make this an opt-in feature through a setting, for those that do have a local users database and can store the hash like this?

moggers87:

PR #258, opened 2018-02-22
Fix the wording in 2FA disable template

Wolf480pl:
NOTE: I meant to open an issue about the wording, but figured out providing an example diff would be more helpful. Please don't be angry if the PR doesn't meet criteria necessary to merge it.

## Description
Change "compromises your account security" to "weakens your account security"
as disabling 2FA alone does not necessarily cause a security breach.

## Motivation and Context
The wording in the 2FA disable template is misleading.
One usually connects "compromises your account security"  with "security compromise" i.e. a security breach.
It makes it seem like you will get pwned the moment you disable 2FA, which, in most cases, is not true.

## How Has This Been Tested?
This is a simple change in wording, I don't think it needs testing.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/templates/two_factor/profile/disable.html
+++ b/two_factor/templates/two_factor/profile/disable.html
@@ -4,7 +4,7 @@
 {% block content %}
   <h1>{% block title %}{% trans "Disable Two-factor Authentication" %}{% endblock %}</h1>
   <p>{% blocktrans %}You are about to disable two-factor authentication. This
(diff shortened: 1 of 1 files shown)


Bouke:
> Change "compromises your account security" to "weakens your account security"
> as disabling 2FA alone does not necessarily cause a security breach.

The original wording was a bit too strong, I agree. Thanks for noticing and putting in the PR!

PR #267, opened 2018-03-25
Email 2fa

Atterratio:
Make possible two-factor authorization using emails.

## Motivation and Context

Although this way of two-factor authentication is not very reliable, it can be preferable for some users.

## How Has This Been Tested?

Tested in a local project.

## Types of changes

## Checklist:


two_factor/views/core.py
```diff
@@ -332,6 +340,12 @@ def get_device(self, **kwargs):
         kwargs['name'] = 'default'
         kwargs['user'] = self.request.user
 
+        if method == 'email':
+            if self.storage.validated_step_data.get('email') and not self.request.user.email:
+                self.request.user.email = self.storage.validated_step_data.get('email').get('email')
```
moggers87: When does this get saved?

two_factor/templates/two_factor/email/email_template.html
```diff
@@ -0,0 +1 @@
+Your authentication token for user {{ user }} is {{ token }}.
```
moggers87: * This should be in `{% blocktrans %}` tags at least.
* It's a bit curt and doesn't say where the email has come from (imagine the user received this message unexpectedly due to a compromised password)

moggers87:
Also, I see you've opened and closed multiple copies of this pull request. If you want to make changes you can just push to your branch and the pull request is automatically updated - there's no need to re-create it each time.

two_factor/views/core.py
Atterratio: [Here](https://github.com/Atterratio/django-two-factor-auth/blob/1afe5f11a81f01e207b11d31702c3eda0b78ed9f/two_factor/forms.py#L79-L81), after input valid token, its allows add for user only confirmed email.

Atterratio: I'm move it in done method of setup view.

Cabalist:
Hey!  Super interested in seeing this ready to merge.   What is left to happen?

moggers87:
it needs re-reviewing and conflicts need to be fixed.

WayneLambert:
Hi @Atterratio, thank you for putting in the development work to make authentication using emails an option for this package.

I too am keen on using this package for an implementation of another project I'm working on. I have already successfully implemented using just the authenticator app option. 

The challenge in my case is requiring an additional method other than using an authenticator app since a minority of users may not have a smartphone. The SMS or call options are not viable due to budget constraints and therefore an additional email option would be great. Yubikey hardware is also not an option for me.

@Bouke @moggers87 It's been a while since this was last reviewed and I wondered if it was still your intention to merge and publish this change? If so, what is the likely timeline for this to happen please?

I'm grateful for your contributions to the development and maintenance of this package.

pickfire:
Since there has not been change since quite some time ago. I created a patch based on this, I also squashed the commits into one commit. https://github.com/Bouke/django-two-factor-auth/pull/423

claudep:
Superseded by #423.

PR #272, opened 2018-06-14
Update django-phonenumber-field dep and tox deps

craigloftus:
Bumping the requirements to `django-phonenumber-field<2.99` and updating the deps in tox. I removed `phonenumbers` from the tox deps as that is no longer a direct dependency.

I've run this with tox locally with the exception of the py34 environment.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -42,7 +42,7 @@ deps =
     coverage
 
     django-formtools
-    django-phonenumber-field>=0.7.2,<0.99
+    django-phonenumber-field>=1.1.0,<2.99
     django_otp>=0.3.4,<0.99
     phonenumbers>=7.0.9,<7.99
     qrcode>=4.0.0,<6.99
--- a/setup.py
+++ b/setup.py
@@ -15,8 +15,9 @@ setup(
         'Django>=1.11',
         'django_otp>=0.3.4,<0.99',
         'qrcode>=4.0.0,<6.99',
-        'django-phonenumber-field>=1.1.0,<1.99',
+        'django-phonenumber-field>=1.1.0,<2.99',
         'django-formtools',
+        'phonenumbers>=7.0.9,<7.99',
     ],
     extras_require={
         'Call': ['twilio>=6.0'],


tox.ini
```diff
@@ -39,9 +39,8 @@ deps =
     coverage
 
     django-formtools
-    django-phonenumber-field>=0.7.2,<0.99
+    django-phonenumber-field>=0.7.2,<2.99
```
moggers87: Shouldn't this match what's in setup.py?

craigloftus: Apologies, I had overlooked the lower bounds.

tox.ini
```diff
@@ -39,9 +39,8 @@ deps =
     coverage
 
     django-formtools
-    django-phonenumber-field>=0.7.2,<0.99
+    django-phonenumber-field>=1.1.0,<2.99
     django_otp>=0.3.4,<0.99
-    phonenumbers>=7.0.9,<7.99
```
Bouke: `phonenumbers` is a direct dependency, see also: https://github.com/Bouke/django-two-factor-auth/blob/0ec439604abebfd087f3bb282ec8cc2333777b29/two_factor/templatetags/two_factor.py#L3

craigloftus: Ah. I have updated `setup.py` to include the dep then, rather than dropping it from `tox.ini`.

Bouke:
We install the phonenumbers package in `tox.ini` and not in `setup.py`, because you can either install `phonenumbers` or `phonenumberslite`. The latter being a smaller database than the former. 

> The phonenumberslite version of the package does not include the geocoding, carrier and timezone metadata, which can be useful if you have problems installing the main phonenumbers package due to space/memory limitations.

When we add a hard dependency on `phonenumbers`, users can no longer opt to choose the lite dataset. See also https://github.com/Bouke/django-two-factor-auth/issues/213.

However... once [this commit](https://github.com/stefanfoulis/django-phonenumber-field/commit/fa21f7731ec191c6f20d3bef08631883c6a2eb78) gets released, that package always hard depends on `phonenumbers`, so this no longer holds true. Maybe we should do something similar, what do you think?

craigloftus:
The question is somewhat moot, as once `django-phonenumber-field` is updated then users of `django-two-factor-auth` will get the full `phonenumbers` package anyway.

> However... once this commit gets released, that package always hard depends on phonenumbers, so this no longer holds true. Maybe we should do something similar, what do you think?

I suggested a direct dependency on `phonenumbers` in that case (and this one) because it was the simple choice, and the 'cost' doesn't seem all that bad.

The [memory footprint](https://github.com/daviddrysdale/python-phonenumbers#memory-usage) should be the same, the on disk size is 15M vs 50M, and download is 400K vs 3M (for wheels).

The long term solution is to get `phonenumbers` to use environment markers to allow the extras to be installed at the project level when needed.

ryancausey:
It looks like in [this PR](https://github.com/stefanfoulis/django-phonenumber-field/pull/236), which appears released in version 2.1.0, django-phonenumber-field moves the dependency to extras_require and requires the developer to choose to install one or the other. Maybe django-two-factor-auth can take a similar approach?

For what it's worth, I've been using phonenumberslite and it doesn't appear to me that the usage in templatetags/two_factor.py specifically requires one or the other. I could be wrong, however.

moggers87:
@ryancausey that's probably the best route to go (other than fixing phonenumbers' awful packaging)

We'd need a documentation update as well as updating the version of django-phonenumber-field that we require.

ryancausey:
Not sure if this was the right way to do it, but @craigloftus I [created a PR](https://github.com/craigloftus/django-two-factor-auth/pull/1) against your fork that should do the trick. I also merged master into it since github was complaining that the branch was out of date. Let me know if you have feedback.

craigloftus:
@ryancausey I'd suggest you make a fresh PR and just reference this one in the description.

PR #278, opened 2018-08-02
Opt token are digits, not integers

lmignon:
Indeed, generated tokens can start with '0'. If we use an IntegerField into the form, the 0 will desapear and the value becomes invalid

## Description

Replace the IntergerField by RegexField for otp_token into AuthenticationTokenForm

## Motivation and Context

 generated tokens can start with '0'. If we use an IntegerField into the form, the 0 will desapear and the value becomes invalid

## How Has This Been Tested?
Manual test: Tested by entering valid tokens starting with '0' and invalid tokens (alpha, length< totp_digits, length> totp_digits, ...)

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/forms.py
+++ b/two_factor/forms.py
@@ -139,9 +139,10 @@ class DisableForm(forms.Form):
 
 
 class AuthenticationTokenForm(OTPAuthenticationFormMixin, Form):
-    otp_token = forms.IntegerField(label=_("Token"), min_value=1,
-                                   max_value=int('9' * totp_digits()))
-
+    otp_token = forms.RegexField(label=_("Token"),
+                                 regex=r'^[0-9]*$',
+                                 min_length=totp_digits(),
+                                 max_length=totp_digits())
     otp_token.widget.attrs.update({'autofocus': 'autofocus'})
 
     # Our authentication form has an additional submit button to go to the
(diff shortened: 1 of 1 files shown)


moggers87:
I'm not convinced that this *is* a bug - the number of required digits is static as I understand it. For example, a TOTP device/app will always return the same number of digits meaning there's no difference between `int('0072')` and `int(72)`.

lmignon:
@moggers87 you're probably right but user complains that when they receive a generated token starting with '0', the '0' is removed when the form is submitted and the validation fails for the given token.

Bouke:
I think that with this change, mobile devices won't show the on-screen number keyboard.

moggers87:
@lmignon ah, sorry I misunderstood what the issue was.

I've not tested this claim, but according to [this stackoverflow answer](https://stackoverflow.com/a/9318515) it's possible to convince mobile browsers to only display the number pad on a text field using the `pattern` attribute. Does the `RegexField` populate that attribute or is it just a backend check?

lmignon:
@moggers87 
>  Does the RegexField populate that attribute or is it just a backend check?

Yes It's just a backend check.

Bouke:
The referenced answer from SO works as advertised on my iPhone. Not sure about Android, can anyone verify this?

I've updated the code according to this:
```
    otp_token = forms.RegexField(label=_("Token"),
                                 regex=r'^[0-9]*$',
                                 min_length=totp_digits(),
                                 max_length=totp_digits())
    otp_token.widget.attrs.update({
        'autofocus': 'autofocus',
        'pattern': '[0-9]*', # hint to show numeric keyboard for on-screen keyboards
    })
```

img_623599ddc065-1

claudep:
@lmignon, do you think you could rebase your patch and add a test?

claudep:
Patch rebased including tests in #479.

PR #284, opened 2018-11-12
Adds middleware to enforce 2fa

singh1114:
## Description

As far as I know, The current 2FA enforcement involves admin intervention. With the following change, people can directly include middleware in the settings file and things will work out of the box.

## Motivation and Context

We are using this software for in our django app. We wanted to enforce 2FA. But the existing solution doesn't allowed the free movement and stopped the user from moving forward if the user has not enabled 2FA. 

We wanted it differently, we wanted the users to move forward but only to the urls where the user can either `logout` or enable 2FA.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/middleware/enforce2fa.py
+++ b/two_factor/middleware/enforce2fa.py
@@ -1,0 +1,21 @@
+"""Middleware to check if staff user has 2FA unabled."""
+
+from django.shortcuts import redirect
+from django.urls import resolve, reverse
+
+
+class Enforce2FAMiddleware:
+    """Enforce 2FA middleware."""
+    def __init__(self, get_response):
+        """Initialize the middleware."""
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """Wrap around actual Admin calls."""
+        response = self.get_response(request)
+        if resolve(request.path).app_name == 'two_factor' or (
+                resolve(request.path).url_name == 'logout'):
+            return response
+        if not request.user.is_verified():
+            return redirect(reverse('two_factor:profile'))
+        return response


singh1114:
@Bouke Please check if the following change is required or not. If yes, Please share if you want any modifications as well.

two_factor/middleware/enforce2fa.py
```diff
@@ -0,0 +1,21 @@
...
+    """Enforce 2FA middleware."""
+    def __init__(self, get_response):
+        """Initialize the middleware."""
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """Wrap around actual Admin calls."""
+        response = self.get_response(request)
+        if resolve(request.path).app_name == 'two_factor' or (
+                resolve(request.path).url_name == 'logout'):
```
moggers87: How do we know the logout URL is called logout?

two_factor/middleware/enforce2fa.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Enforce2FAMiddleware:
+    """Enforce 2FA middleware."""
+    def __init__(self, get_response):
+        """Initialize the middleware."""
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """Wrap around actual Admin calls."""
```
moggers87: This code won't limit itself to just the admin site as it makes no such check. Either code needs to do what the comment says or the comment needs changing.

two_factor/middleware/enforce2fa.py
```diff
@@ -0,0 +1,21 @@
+"""Middleware to check if staff user has 2FA unabled."""
+
+from django.shortcuts import redirect
+from django.urls import resolve, reverse
+
+
+class Enforce2FAMiddleware:
```
moggers87: This package still supports Python 2.7 for Django 1.11, so that should be a subclass of `object`.

two_factor/middleware/enforce2fa.py
singh1114: @moggers87 Is there any way of knowing this?

moggers87: Not that I know of.

PR #293, opened 2019-03-14
Add QueryableFake gateway

christopherdcunha:
## Description

This change adds a fake gateway that can be queried programatically.

## Motivation and Context

This allows my functional tests to receive tokens and complete the 2FA flow.

## How Has This Been Tested?

This has been tested via the unittests in this PR and a functional test similar to the one in the docstring of the QueryableFake class.

### Documentation `configuration.html#general-settings`

image

### Documentation `configuration.html#queryablefake-gateway`

image

## Types of changes

## Checklist:


docs/configuration.rst
```diff
@@ -19,6 +19,8 @@ General Settings
     Twilio_.
   * ``'two_factor.gateways.fake.Fake'``  for development, recording tokens to the
     default logger.
+  * ``'two_factor.gateways.fake.QueryableFake'``  for testing, recording tokens
+    to the ``QueryableFake`` singleton.
```
moggers87: The new gateway is not a singleton. I know what you're trying to say here (`sms_tokens` and `call_tokens` are initialised when the class is created not when an object is instantiated) but "singleton" would mean that  things like `QueryableFake().sms_tokens = {}` would affect other instances of `QueryableFake` because there would only be one instance of the class.

tests/test_gateways.py
```diff
@@ -117,3 +117,15 @@ def test_gateway(self, logger):
             fake.send_sms(device=Mock(number=PhoneNumber.from_string('+123')), token=code)
             logger.info.assert_called_with(
                 'Fake SMS to %s: "Your token is: %s"', '+123', code)
+
+
+class QueryableFakeGatewayTest(TestCase):
```
moggers87: More tests please. E.g. do `call_tokens` and `sms_tokens` get reset properly after each testcase?

christopherdcunha:
Thanks for the review @moggers87 - I've added commit cfc695d which addresses the issues you mentioned (or tries to at least). If you are happy with the changes, then I'll squash the commit and push.

docs/configuration.rst
```diff
@@ -28,6 +30,8 @@ General Settings
     Twilio_.
   * ``'two_factor.gateways.fake.Fake'``  for development, recording tokens to the
     default logger.
+  * ``'two_factor.gateways.fake.QueryableFake'``  for testing, recording tokens
```
Bouke: If this is merged, the text "Currently two gateways are bundled:" no longer holds.

Bouke:
An alternative would be to just use fixed tokens instead. This would simplify the tests, while allowing the same use case. For example, it could work e.g. like this:

* 000000 -> valid token
* 000001 -> invalid token

christopherdcunha:
Thanks for the review @Bouke 

RE: The fixed tokens approach.

I think that you're suggesting an approach where [`PhoneDevice.verify_token(self, token)`](https://github.com/Bouke/django-two-factor-auth/blob/87b644ce5bbd9f65959bd619b1782868f14888d3/two_factor/models.py#L87-L99) will simply return `token == '000000'`. And a test case would simply check if [`gateways.fake.Fake.send_sms(device, token)`](https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/gateways/fake.py#L33-L35) was called.

Is my understanding correct?

dopry:
@Bouke @christopherdcunha,  is this still something worth pursuing? It sounds like there may need to be some consensus about whether this is the best approach to testing.

christopherdcunha:
> @Bouke @christopherdcunha, is this still something worth pursuing? It sounds like there may need to be some consensus about whether this is the best approach to testing.

Agreed. I personally prefer/use the approach I've suggested here, but I'm totally happy to close this PR if it doesn't make sense for the broader "library user base".

PR #296, opened 2019-04-15
Only import 'patch_admin' module if TWO_FACTOR_PATCH_ADMIN==True

spookylukey:
Allow `two_factor` to be used if `admin` is not in `INSTALLED_APPS`.

Without this patch, you get `LookupError: No installed app with label 'admin'` if you import `two_factor`, whatever the value of `TWO_FACTOR_PATCH_ADMIN`. Should be fairly self explanatory!

I didn't add any tests since this kind of import error is hard to test.

Thanks!

### DIFF
--- a/two_factor/apps.py
+++ b/two_factor/apps.py
@@ -7,6 +7,6 @@ class TwoFactorConfig(AppConfig):
     verbose_name = "Django Two Factor Authentication"
 
     def ready(self):
-        from .admin import patch_admin
         if getattr(settings, 'TWO_FACTOR_PATCH_ADMIN', True):
+            from .admin import patch_admin
             patch_admin()


moggers87:
I don't have this issue myself, could you open an [issue](https://github.com/Bouke/django-two-factor-auth/issues/new) and provide some more information please?

spookylukey:
> I don't have this issue myself, could you open an [issue](https://github.com/Bouke/django-two-factor-auth/issues/new) and provide some more information please?

See #297

moggers87:
Codecov is preventing me from merging this, so I'm going to go ahead and dot it manually. The changes here are not that easy to unittest and it's a fairly significant break for Django 2.1+ users.

moggers87:
@Bouke is it possible for you to override the codecov checks?

Bouke:
Sure! And thanks for the effort 👍 .

PR #301, opened 2019-05-08
Fix error PhoneDevice in Django 2.2

### LINKED ISSUE #298: Compatibility issues with django 2.2
## Problem
I have been using two_factor with Django 2.1 for a while without any problems (thank you for all your work!). However, I recently upgraded to Django 2.2 and am having some compatibility issues. Whilst I am still able to create ```PhoneDevice``` objects, I can now no longer delete them (both within the shell environment and the relevant views e.g. ```PhoneDeleteView``` and ```DisableView```). In both cases, I get the same error. For example, if I enter the following in the shell ``` PhoneDevice.objects.first().delete() ``` I get the following error:

``` console
  File "<console>", line 1, in <module>
  File "/home/iwilliams/.virtualenvs/twofactor/lib/python3.6/site-packages/django/db/models/base.py", line 918, in delete
    collector.collect([self], keep_parents=keep_parents)
  File "/home/iwilliams/.virtualenvs/twofactor/lib/python3.6/site-packages/django/db/models/deletion.py", line 192, in collect
    reverse_dependency=reverse_dependency)
  File "/home/iwilliams/.virtualenvs/twofactor/lib/python3.6/site-packages/django/db/models/deletion.py", line 94, in add
    if obj not in instances:
TypeError: unhashable type: 'PhoneDevice'
```

## Possible Solution
Part of the problem seems to lie with the following method:

``` Python
class PhoneDevice(Object):
   ...
    def __eq__(self, other):
        if not isinstance(other, PhoneDevice):
            return False
        return self.number == other.number \
            and self.method == other.method \
            and self.key == other.key
```
When the above function is commented out, I am able to remove PhoneDevice objects. Perhaps on a related note, I should note that I can add ```TOTPDevice``` and ```YubikeyDevice``` instances in Django 2.2 without a problem, both of which don't override the base ```__eq__``` method. 

Do you have any further insight into what is causing this problem/ how it can be properly resolved?

Thanks in advance!

## Environment Info
* Python version: 3.6.7
* Django version: 2.2
* django-otp version: 0.6.0
* django-two-factor-auth version: 1.8.0

g-clef:
## Description

Adding a `__hash__` function to the `PhoneDevice` model object, to allow it to be used in hash-based searches like `obj in set_of_objects`.

## Motivation and Context
Removing a phone device (e.g. deactivating 2-factor from an account that has a phone device attached) causes a 500 error right now in django 2.2 because django tries to see if the model object is in a set. Since the PhoneDevice model object overrides `__eq__`, it needs a `__hash__` function to allow it to be searched in a set. This resolves #298.

For the actual Hash value I used a hash of a format string made from the repr's of the number, key, and method fields.

## How Has This Been Tested?
All tests pass. Also I installed this on a failing 2.2 Django installation and confirmed that the error with removing a Phone device is resolved.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_views_phone.py
+++ b/tests/test_views_phone.py
@@ -55,7 +55,13 @@ class PhoneSetupTest(UserMixin, TestCase):
         self.assertContains(response, 'We\'ve sent a token to your phone')
         device = response.context_data['wizard']['form'].device
         fake.return_value.make_call.assert_called_with(
-            device=device, token='%06d' % totp(device.bin_key))
+            device=mock.ANY, token='%06d' % totp(device.bin_key))
+
+        args, kwargs = fake.return_value.make_call.call_args
+        submitted_device = kwargs['device']
+        self.assertEqual(submitted_device.number, device.number)
+        self.assertEqual(submitted_device.key, device.key)
+        self.assertEqual(submitted_device.method, device.method)
 
         response = self._post({'phone_setup_view-current_step': 'validation',
                                'validation-token': '123456'})
--- a/two_factor/models.py
+++ b/two_factor/models.py
@@ -80,16 +80,6 @@ class PhoneDevice(Device):
             self.method,
         )
 
-    def __eq__(self, other):
-        if not isinstance(other, PhoneDevice):
-            return False
-        return self.number == other.number \
-            and self.method == other.method \
-            and self.key == other.key
-
-    def __hash__(self):
-        return hash("{!r},{!r},{!r}".format(self.number, self.method, self.key))
-
     @property
     def bin_key(self):
         return unhexlify(self.key.encode())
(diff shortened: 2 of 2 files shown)


moggers87:
The only reason `PhoneDevice.__eq__` was ever added was to make tests pass. No other device model implements its own `__eq__` method and it was only added because Django 1.7 changed the behaviour around unsaved model objects equalling one another. No doubt we'll have to add *another* fix in a few years when something else changes.

I think a better idea would be to fix the `tests.test_views_phone.PhoneSetupTest.test_setup` test (i.e. just check the object properties are the same) and remove the `__eq__` method all together.

g-clef:
Okay. I just pushed up that change (removing both `__eq__` and `__hash__`, and fixing the unit tests to check equality by testing object properties).

moggers87:
Looks good, I'm still waiting for #300 to be merged so we can be sure these changes don't break anything :smile:

PR #294, opened 2019-03-26
Update django-phonenumber-field and let user decide which phonenumbers lib to use

ryancausey:
This is based on the discussion and work on #272

I added `phonenumbers` and `phonenumberslite` as extras_require. I've also updated the install documentation to note the need to add one of the two libraries. I also bumped the phonenumbers version as it is on 8.x now.

I ran all of the tests for python 2.7 and 3.6 using `tox` and all passed. I am having issues figuring out how to get `tox` going for the other versions, so I haven't yet tested them. I also manually confirmed I could install using the extras syntax and get the correct extra, be it `phonenumbers` or `phonenumberslite`.

PR #313, opened 2019-08-05
update django-phonenumber-field dependency version and allow picking between phonenumbers and phonenumberslite

v1kku:
Same as #294 but with the tests fixed and max version of django-phonenumber-field set to 3.99

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -8,6 +8,19 @@ and its dependencies:
 
     $ pip install django-two-factor-auth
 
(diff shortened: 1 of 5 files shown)


moggers87:
@Bouke looks like codecov is stalled again?

Bouke:
> @Bouke looks like codecov is stalled again?

It looks like it yes. If this continues to happen, we might want to remove those status checks for PRs.

Bouke:
Thank you all!

PR #323, opened 2019-12-10
Pin django-formtools for python 2.7 support

CollinEMac:
Pin django-formtools for python 2.7 support.

## Description
When trying to use django-two-factor-auth version 1.8.0 we realized that django-formtools was not pinned. This caused the code to break because the latest version of django-formtools was not compatible with version 1.8.0. In django-formtools >2.0 the super() call is expecting a class argument; however django-two-factor-auth version 1.8.0 is not passing the argument. I don't know how to update the docs on a minor version.

For reference:
https://github.com/jazzband/django-formtools/blob/2.0/formtools/wizard/views.py#L122
https://github.com/jazzband/django-formtools/blob/2.2/formtools/wizard/views.py#L122

It was producing this error:
```
backend_1_5b82166a2320 |   File "/usr/local/lib/python2.7/site-packages/two_factor/urls.py", line 11, in <module>
backend_1_5b82166a2320 |     view=LoginView.as_view(),
backend_1_5b82166a2320 |   File "/usr/local/lib/python2.7/site-packages/formtools/wizard/views.py", line 122, in as_view
backend_1_5b82166a2320 |     return super().as_view(**initkwargs)
backend_1_5b82166a2320 | TypeError: super() takes at least 1 argument (0 given)
```

## Motivation and Context
We're using Python 2.7 and thus we need to use django-two-factor-auth version 1.8.0. We're in the process of upgrading to python 3.7 but we need this temporary fix while we upgrade.

## How Has This Been Tested?
After making this change, reinstalled dependencies, and started docker the error went away.

## Types of changes

## Checklist:

**NOTE:** once this PR is approved, tag 1.8.1 should be created off of this branch in lieu of merging.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
         'django_otp>=0.3.4,<0.99',
         'qrcode>=4.0.0,<6.99',
         'django-phonenumber-field>=1.1.0,<1.99',
-        'django-formtools',
+        'django-formtools<=2.0',
     ],
     extras_require={
         'Call': ['twilio>=6.0'],
--- a/tox.ini
+++ b/tox.ini
@@ -41,7 +41,7 @@ deps =
     yubikey: django-otp-yubikey
     coverage
 
-    django-formtools
+    django-formtools<=2.0
     django-phonenumber-field>=0.7.2,<0.99
     django_otp>=0.3.4,<0.99
     phonenumbers>=7.0.9,<7.99


claudep:
If you do that, you will also have to limit the Django version, as django-formtools <2 does not support Django 3.0.

moggers87:
It's trivial for users to limit the version of django-formtools themselves:

```
# requirement.txt
django>=1.11,<2
django-two-factor-auth
django-formtools<=2
```

Is there any reason you can't do this?

Also, is 1.8.0 a typo? Version 1.9.1 was released this summer and supports both Python 2.7 and Django 1.11

CollinEMac:
> It's trivial for users to limit the version of django-formtools themselves:
> 
> ```
> # requirement.txt
> django>=1.11,<2
> django-two-factor-auth
> django-formtools<=2
> ```
> 
> Is there any reason you can't do this?
> 
> Also, is 1.8.0 a typo? Version 1.9.1 was released this summer and supports both Python 2.7 and Django 1.11

I didn't realize we could pin django-formtools in our own codebase. Thanks. I'm closing this PR.

PR #325, opened 2019-12-11
Mentioned Django 3.0 support at various places

claudep:


README.rst
```diff
@@ -35,7 +35,7 @@ user. Although the package is optional, it improves account security control
 over ``django.contrib.sessions``.
 
 Compatible with modern Django versions. At the moment of writing that's
-including 1.11, 2.1, and 2.2 on Python 3.4, 3.5, 3.6 and 3.7. Documentation is
+including 1.11, 2.1, 2.2, and 3.0 on Python 3.4 and up. Documentation is
```
moggers87: We don't actually have 3.8 tests yet, nor any future 3.x release. IMO it's better to be explicit about what versions we test against.

PR #326, opened 2019-12-12
Allow django-phonenumber-field 4.x

Natureshadow:
According to changelog, the version bump was due to a change in handling of invalid phone numbers, that should not break anything existing.

## Types of changes

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         'Django>=1.11',
         'django_otp>=0.6.0,<0.99',
         'qrcode>=4.0.0,<6.99',
-        'django-phonenumber-field>=1.1.0,<3.99',
+        'django-phonenumber-field>=1.1.0,<4.99',
         'django-formtools',
     ],
     extras_require={


moggers87:
What's the change exactly? That's a new major version, so clearly upstream think it's a big deal.

claudep:
Changelog for 4.0:
`The big version bump is due to the change in how invalid phone numbers are handled. Starting with 2.4.0 we added very aggressive validation, which raised ValueError for invalid numbers. This caused problems in unexpected places (like when filtering a queryset). Starting with 4.0.0 we acknowledge that we can not completely prevent invalid numbers from entering the system. Changes directly to the database, validation changes in the upstream phonenumbers library, changes in the django settings may all lead to invalid numbers in the database. Now it is possible to save an invalid number to the database and __str__ and __repr__ clearly indicate invalid numbers.`

Natureshadow:
@moggers87 Only if downstream relies on exceptions for normal control flow. If you catch ValueError to validate phone numbers instead of using a validator, this is bad news. But it's your own fault then ;).

In all other cases, it just reacts less indignant to outside changes to the database.

moggers87:
OK, but is that the case for django-two-factor-auth? You say it "should not" break anything, but have you actually tested this?

Natureshadow:
I assume the current test suite does already verify that invalid phone numbers are caught correctly. If not, it is out of the scope of this change.

Bouke:
Any breaking changes resulting from this change are of course within the scope of this change. So that needs to be verified; preferably by the one proposing the change.

claudep:
Superseded by #443

PR #328, opened 2020-01-03
Add a registry to dynamically (un)register auth methods

claudep:
This is a tentative patch to add *some* plugin architecture to allow for modular auth method implementations, following comments on PR #215. It's rather minimal and definitely not totally modular yet, but at least I have a proof of concept patch on my local system to add the U2F functionality (which I hope to integrate to this package).

I'm totally open to improve it following your comments.


moggers87:
I've not had a chance to do a proper review, but how does this compare to https://github.com/Bouke/django-two-factor-auth/pull/280 ?

claudep:
Crap, didn't see that one. However, I'm sure we can go somewhere by uniting ideas. What I called `methods`, @Bouke called it `plugins`. Do you think `plugins` is somewhat more standard as extension conceptual name?

claudep:
This is obsolete now we have a plugin architecture in place.

PR #329, opened 2020-01-04
Used trimmed blocktrans option to avoid garbage newlines for translators

claudep:


Bouke:
While this change in itself is fine, the result is that a lot of the translations now dropped from 100% to 67%.

claudep:
Sure, so poking translators to complete them before the next release would be nice. Doing it currently for French.

PR #280, opened 2018-08-10
[WIP] Pluggable providers

Bouke:
Work in progress for pluggable factor providers.

## Motivation and Context
The issue to add additional factor provider comes up frequently. We currently have questions / PRs for adding U2F and e-mail factor providers. Currently this doesn't compose, as the setup / login views become overly complex. This rewrite attempts to improve the situation.

## Description
Factor providers can be added as Django apps. Implementers are expected to implement a contract. See the generator / phone number providers as example implementations. 

This PR is a very rough draft. There are some areas that need further improvement;
- the views becoming even more complex, due to all the callbacks on the contract
- inconsistency in how forms are configured / devices are created, perhaps through a unified interface

There's also some features missing currently:
- ability to configure backup devices 

Some things to resolve:
- handle db migrations

### Extensibility
Both builtin provider implementations contain a subclass of AppConfig. This subclass will be the entry point for further extension. Users of these providers can provide their custom subclass to overwrite the defaults and provide additional functionality.

## How Has This Been Tested?
- PR is not incomplete.
- Unit tests are broken.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #331, opened 2020-01-04
Moved some capabilities to plugins folder

claudep:
Based on #280 and my recent work on making a method registry in #328, this is a first step in that direction, trying to isolate code per auth method. That's not the end of the road, but this commit should at least pass the tests.

We are breaking compatibility because we moved the `PhoneDevice` models to `plugins/phonenumber` sub-app, but I think that's a price to pay for added modularity.


moggers87:
Looks good! Just a couple of questions:

1. Is Django going to happy about migrations for `PhoneDevice` being in a different app?
2. How much more work would be needed to make `django-phonenumber-field` optional? From what I can see, this is mostly done with the exception of migrations.

claudep:
1. The fact that tests are running fine makes me think the application of migrations is OK (thanks to the `Meta.app_label` I guess). When doing a `makemigrations`, the migration file is created in `two_factor/migrations`, so they will need to be moved by hand (no impact on users IMHO).

2. Yes, this is a good step in that direction. Would you like this in this same PR?

claudep:
I just pushed one more commit to go towards phonenumbers independence. Still not there yet.

moggers87:
Making `django-phonenumber-field` optional is not required for this PR to be merged. I was just getting an idea of how much future work would be required if we wanted to go fully modular with this package.

claudep:
@moggers87, @Bouke, I would really like to make some progress on this (with the ultimate goal of adding u2f support). Do you think you will have some time for the reviews during the next weeks?

Bouke:
Overall it looks pretty complete. However where did the views and templates for managing phone devices go?

claudep:
The views are in `two_factor/plugins/phonenumber/views.py`. I did not move the phone_register.html template yet. If you think it would be better to do it in this patch, I can do it.

docs/class-reference.rst
```diff
@@ -46,21 +46,16 @@ Signals
    ``request``
        The ``HttpRequest`` in which the user was verified.
 
-Template Tags
--------------
-.. automodule:: two_factor.templatetags.two_factor
```
Bouke: Are those now documented elsewhere?

Bouke:
So when merging this PR, is there any difference in functionality? Are there any known regressions? Does the changing of models result in data-loss, or weirdness with (generating) the migrations? -- Is it "safe" to merge this PR, or are there sharp edges?

docs/class-reference.rst
claudep: They are not used as template tags any longer, they have been moved to ` two_factor/plugins/phonenumber/utils.py`. We should check if this one will appear in docs now.

Bouke: Can we keep making those available as template tags? Users might depend on them in their (custom) views. Or is there reason to drop them?

claudep:
In my opinion, we are just moving things here, without any change in functionality. Of course, moving things can always break imports from other packages.

docs/class-reference.rst
claudep: Would you like to keep them at the current import path? In the spirit of the PR, they should be at least moved to `two_factor.plugins.phonenumber.templatetags.phonenumber`.

moggers87: I see no reason not to move them - the name used to import them in templates would remain the same.

claudep: With my proposal, the `{% load two_factor %}` would be `{% load phonenumber %}`. Acceptable change?

moggers87: Ah, I misread your comment. Is it really necessary to rename the template library as well as move it? It's the only template tag library this package implements and we can always change it later if we really need to.

claudep: The current template tags are all related to the phonenumber functionality. That's why it would cleaner to have them inside `plugins/phonenumber`. I don't know if keeping an empty `two_factor.templatetags.two_factor` file makes sense. And if you would rather don't touch them at all, that's your choice.

claudep: So what do you decide about this templatetags question?

moggers87: (sorry, i've been a bit busy this week)

My thinking was to just move `two_factor.templatetags.two_factor` to `two_factor.plugins.phonenumber.templatetags.two_factor`. That way templates won't have to be changed and users who want phone devices on their install will already be adding `two_factor.plugings.phonenumber` to the settings.py anyway. IMO, the less changes users have to make because of this PR, the fewer bug reports we'll get later.

I don't think we need to leave an empty `two_factor.templatetags.two_factor` because our template library doesn't contain anything that's not related to phone numbers (as you pointed out).

claudep: The problem I see with this solution is that as soon as we'll need a template tag in the core namespace, we won't able to use `two_factor.templatetags.two_factor` because of the namespace conflict.

moggers87: OK, let's rename it then :smile_cat:

claudep:
The failure is a connection timeout while fetching Python 3.5

claudep:
Are there still reservations about going forward with this patch?

moggers87:
I still need to do a code review, but I thought you were still making changes. As that's not the case I'll try and get on with that this weekend.

To be clear: I like this PR and I think it's the right way froward. The code review will just be me nit-picking :cat:

claudep:
Feel free to nit-pick as much as you want!

docs/installation.rst
```diff
@@ -39,6 +39,7 @@ Add the following apps to the ``INSTALLED_APPS``:
         'django_otp.plugins.otp_static',
         'django_otp.plugins.otp_totp',
         'two_factor',
+        'two_factor.plugins.phonenumber',  # <- if you want phone number capability.
```
moggers87: I feel like this could be expanded upon. Maybe mention it with the the `PHONENUMBER_*` settings docs?

docs/installation.rst
```diff
@@ -39,6 +39,7 @@ Add the following apps to the ``INSTALLED_APPS``:
...
     )
```
moggers87: Might be nice to have `two_factor.plugins.yubikey` in here too. I know it's not technically needed, but it does advertise the fact that we support yubikey.

moggers87:
I have yet to test this out on a real Django app and while I could probably get a Twilio working on my app (Twilio has a free trial) I don't own a Yubikey. @claudep how much testing of your own have you done?

claudep:
For me, this is a step preparing the addition of a U2F plugin. I have tested that (still local) plugin with my Yubikey an it's working without breaking the "normal" TOTP functionality. Frankly, I don't know how I can use my Yubikey to make the current yubikey backend working.

After this and the U2F plugin are in, you might release a pre-version so more people are able to test it in real projects. A new major version number can also signify users that major stuff have been "touched".

docs/installation.rst
claudep: Added

docs/installation.rst
claudep: I separated phone-related settings docs with an introductory sentence. Do you find this sufficient?

beckedorf:
Great approach!

Maybe, if we change the structure of this package, we should separate everything related to yubikey and phonenumbers into the plugin folder (probably even regarding TOTP too). If we add a documented interface for the plugins, it becomes a lot easier to include additional devices and reduce special logic for some devices in the core package. Instead we might include some abstract classes and templates to reduce boilerplate in the plugins.

I could see that this paves the way for multiple open issues (and support of WebAuthn @compsult ). Moreover, by decoupling the code testing might become easier.

claudep:
I just rebased on latest master. @moggers87, I think it would be great to finish review and do the merge as soon as possible (no rush, of course), so the upcoming email/webauthn/u2f methods could use the new plugin structure. We may also think about starting a 2.x release series to convey the importance of the structural change.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 ### Added
 - Python 3.10 support
 
+### Changed
+- The phone capability moved to a plugins folder, so if you use that capability
+  and want to keep it, you should add `two_factor.plugins.phonenumber` line in
+  your `INSTALLED_APPS` setting.
```
moggers87, line 9: We should also note that the templatetag library has been changed from "two_factor" to "phonenumber"

moggers87 (changes requested):
I have now (finally) smoke tested this PR with an actual Django app!

One tiny tiny change because it tripped me up when I was testing this PR. Other than that, I am satisfied that these changes aren't going to blow everything up.

moggers87 (approved):
Approved

claudep:
Thanks a lot for merging! Just a quick note for the future: In general, when I keep separated commits in a pull request, that's because it makes sense to cleanly separate them in consistent chunks, so it should be nice to keep the different commits when merging (no squashing). I also could have specified that earlier... so I'm also to blame :-)

PR #334, opened 2020-02-01
Make codecov informational only

moggers87:
## Description

Update Codecov configuration so that it stops blocking PR merges

## Motivation and Context

Codecov is unable to tell the difference between "the developer removed
lines that had coverage" and "the developer added lines without
coverage". Often maintenance work is the former rather than the latter,
but pull requests still end up being marked as "failed".

Codecov should still submit coverage reports as comments to pull
requests.

## How Has This Been Tested?

No.

## Types of changes

## Checklist:

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,5 +4,8 @@ coverage:
       default: false
       tests:
         paths: tests
+        informational: true
       two_factor:
         paths: two_factor
+        informational: true
+    patch: off


claudep:
++1 !

Bouke:
I feel like this change is a workaround and might resolve in missed coverage in the future. Otoh having PRs blocked because of CI failures is also not ideal. So I’m fine with merging this, but would prefer a better solution.

moggers87:
IMO the only way to avoid this is to have 100% code coverage in the first place.

PR #342, opened 2020-03-03
Convert extras names to lowercase

ryan-copperleaf:
## Description

Converts the extra names in the package's `extras_require` to lowercase, working around pypa/pip#4617.

## Motivation and Context

For example, even with the latest pip, `pip install django-setup-tools[SMS]` will not install `twilio`, even though it should.

## How Has This Been Tested?

Tested by installing `.[sms]`, and by confirming that installing `.[SMS]` continues to do what it did before.

N/A

## Types of changes

## Checklist:

- [X] My code follows the code style of this project.
- [ ] ~~My change requires a change to the documentation.~~ (Only `phonenumbers` and `phonenumberslite` seem to be mentioned, which are already lowercase.)
- [ ] ~~I have updated the documentation accordingly.~~
- [ ] ~~I have added tests to cover my changes.~~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,9 +19,9 @@ setup(
         'django-formtools',
     ],
     extras_require={
-        'Call': ['twilio>=6.0'],
-        'SMS': ['twilio>=6.0'],
-        'YubiKey': ['django-otp-yubikey'],
+        'call': ['twilio>=6.0'],
+        'sms': ['twilio>=6.0'],
+        'yubikey': ['django-otp-yubikey'],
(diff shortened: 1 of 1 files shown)


moggers87:
Just an FYI for those who are also affected by this: it seems like this is a wheel bug as specifying `--no-binary :all:` with pip or using easy_install solves the issue. Both of those methods treat `extras_require` keys in a case-insensitive way as expected (I tested this on pip 19.1.1 and easy_install from setuptools 41.2.0)

PR #343, opened 2020-03-04
docs: Fix simple typo, wich -> which

timgates42:
There is a small typo in docs/installation.rst.

Should read `which` rather than `wich`.


moggers87:
Thanks!

PR #346, opened 2020-03-25
Avoid a django_otp version > 0.7.5

jibaku:
related to the use of random_hex_str (https://github.com/Bouke/django-two-factor-auth/blob/master/two_factor/models.py#L57)

## Description

Fix the needed version of django_otp to <= 0.7.5 (to avoid changes made in >=0.8)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
         'Django>=1.11',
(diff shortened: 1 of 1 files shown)


claudep:
Did you install the latest django-two-factor-auth?

jibaku:
Yes but it was not working, I've removed the virtualenv, recreated it and now it's working. Sorry for the inconvenience.

PR #349, opened 2020-03-31
Unifying LoginView API with Django Project (adding support for success_url_allowed_hosts and redirect_authenticated_user) (Revised)

beckedorf:
## Description

Adding `success_url_allowed_hosts` and `redirect_authenticated_user` like in  django.contrib.auth.views.LoginView to `two_factor.views.LoginView`.

## Motivation and Context

The [LoginView in the main Django Project](https://github.com/django/django/blob/4f61810751751b8c5070ce038ea57e949650e9e3/django/contrib/auth/views.py#L40) can be customized using some class variables that are [mentioned in the Django documentation](https://docs.djangoproject.com/en/3.0/topics/auth/default/#all-authentication-views).

`redirect_field_name` and `template_name` of these customizations are already work with `two_factor.views.LoginView`.

I added `success_url_allowed_hosts` and `redirect_authenticated_user`.

These features facilitate to move from the Django authentication system to django-two-factor-auth if developers already rely on some customizations, is a step to unify the public API of this repo and the Django authentication system and (obviously) enriches the functionality of django-two-factor-auth.

## How Has This Been Tested?

Full test coverage is already included.
Other areas of the code should not be affected.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

## References

This reopens and carries on with (closed) pull request https://github.com/Bouke/django-two-factor-auth/pull/348 .

PR #351, opened 2020-04-07
Move login redirect URL logic into its own method

Sjord:
## Description

Move the determination of the redirect URL to its own method, when redirecting in LoginView.done. 

## Motivation and Context

This makes it possible for users to make their own class where `get_redirect_url` is overridden, so that business logic can determine the next URL. For example, [Zulip](https://github.com/zulip/zulip/blob/master/zerver/views/auth.py#L700-L708) has user-specific "realms", so they want to determine the redirect URL only after login.

This moves the redirect URL logic to `get_redirect_url`, and calls that just before calling `redirect`. This moves the URL logic after `signals.user_verified.send`, which means a change in behavior if the URL logic raises an exception. For example, if the given URL cannot be resolved. It used to not call `signals.user_verified.send`, but now it does. I think this is expected behavior; the user is logged in after all.

## How Has This Been Tested?

I added two unit tests to `LoginTest`. I have also tested it in a little example project.

The default `get_redirect_url` shouldn't be able to redirect to an unsafe (off-domain) URL. When overriding `get_redirect_url` it can do whatever it wants, including redirecting to an unsafe URL. This commit adds tests to reflect this.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_views_login.py
+++ b/tests/test_views_login.py
@@ -68,6 +68,26 @@ class LoginTest(UserMixin, TestCase):
              'login_view-current_step': 'auth'})
         self.assertRedirects(response, redirect_url)
 
+    def test_valid_login_with_unsafe_redirect(self):
+        redirect_url = "//example.com"
+        self.create_user()
+        response = self._post({'auth-username': 'bouke@example.com',
+                               'auth-password': 'secret',
+                               'login_view-current_step': 'auth',
+                               'next': redirect_url})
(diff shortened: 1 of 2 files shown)


moggers87:
This looks like it has some overlap with #349 

Is there anything in this PR that's not covered there?

Sjord:
No, it looks like #349 does the same as this PR, and more. So probably better to merge that one and to close this one.

PR #263, opened 2018-03-22
Issue #56 - After successful login with token, use signed cookie to allow login with login step only

compsult:
(scoped to settings.LOGIN_URL). If cookie exists/is valid
on the next login, token steps are skipped

1 - if user checks checkbox to token step "Remember this device for %s days"
2 - create signed cookie with the desired end date 
3 - on next login, if cookie is present/valid, allow login with login step only

## Description
added a method token_required() that returns False when a valid 
signed cookie ('rememberdevice') is present. If not required, methods has_token_step()
and has_backup_step() will return None (skipping that logic).

To enable OTPRequiredMixin() to grant access, if token_required() 
returns False, a device will be added to user

## Motivation and Context

issue #56

## How Has This Been Tested?

yes, tests were added to test_views_login.test_with_backup_phone()
testing both the form change and the cookie logic.

Ubuntu 16.04, Postgres 9.3, Python 3.5.2
Django==2.0.3
django-bootstrap-form==3.4
django-debug-toolbar==1.9.1
django-formtools==2.1
django-otp==0.4.3
django-phonenumber-field==1.3.0
django-two-factor-auth==1.7.0

ran "make test" with no errors

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project. - I tried to :)
- [X] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (I will, if the changes are OK with you)
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

Issue #56, opened 2014-05-22
Remember OTP token for hours


awais786:
I want to implement remember otp for few hours 10 hrs. Means on first attempt admin gives userid/password an then give the OTP and for next 10 hrs when admin try to login again then it asks for userid/password only and skip the OTP till the 10 hrs. 

Whats the best place to implement this in codebase.

shanx:
I would love to have a look at your work! Since I'm looking into implementing this myself, I don't see it in your fork.

danielquinn:
I ended up here as I was trying to figure out how to do this and haven't had any success yet.  Was this issue solved in some other way or abandoned as something the implementing developer should handle?

If it's the latter, how did you lot do this?

compsult:
What do you think about this approach?

1 - Add date column to user_sessions_session called "last_validated"
2 - datestamp  "last_validated" upon a successful token login
3 - Modify OTPRequiredMixin to 
	query user_sessions_session, if user/agent/ip was "last_validated" in {{trusted_days}}
	allow access if authenticated ( without token )

compsult:
Makes sense. 

Made some progress on the signed cookie front but was hoping you might have some ideas about how to handle OTPRequiredMixin. 
The test request.user.is_verified(), which (from my read) requires that request.user.otp_device be populated. Since I am skipping the token logic when there was a recent validated login ,  request.user.otp_device is not getting populated.
Any ideas on how I might solve this?

diegobill:
:+1:

beckedorf:
I just started a new attempt. But still WIP #352

PR #352, opened 2020-04-09
After successful login with token, use signed cookie to allow login with login step only. (Reprise of PR #263 and Issue #56)

beckedorf:
Reprise of https://github.com/Bouke/django-two-factor-auth/pull/263
## Description
During a successful login with a token, the user may choose to remember this browser.
If the same user logs in again on the same browser, a token will not be requested, as the browser
serves as a second factor.

The option to remember a browser is deactived by default. Set `TWO_FACTOR_REMEMBER_COOKIE_AGE` to activate.

The browser will be remembered as long as
- the Cookie that authorizes the browser did not expire,
- the user did not reset his password, and
- the device initially used to authorize the browser is still valid.

## Motivation and Context
Making login easier and save costs, by omitting the second factor when possible (like e.g. Google Login)

## How Has This Been Tested?
Unittests

## Types of changes

## Checklist:


compsult:
@beckedorf you might want to reconsider investing time in this. 
Read about the weaknesses of many 2FA techniques [here](https://blog.trailofbits.com/2019/06/20/getting-2fa-right-in-2019/).
I think webauthn is the future and put together a django demo based on Duo Labs code base [here](https://github.com/compsult/py_webauthn/tree/django_demo). The advantages are spelled out [here](https://webauthn.guide/)

beckedorf:
@compsult webauthn looks very promising. But I think this should be implemented on a lower level (e.g. as a django-otp device) as it is a proper 2nd factor and a lot more complex than a signed cookie. 

Unfortunately it seems like webauthn is not compatible with the current structure/process of django-otp.
(I just started an issue at django-otp https://github.com/django-otp/django-otp/issues/40)

So the question is if it efficient to restructure django-two-factor-auth and django-otp in the long run and get it ready for webauthn?

compsult:
@beckedorf, you're right, it's not compatible with django-otp. Webauthn was developed to replace passwords and the myriad disadvantages of password based authentication. It was put together by the biggest heavy weights (Google, Microsoft, Mozilla). [This article](https://www.twelve21.io/the-case-for-the-web-authentication-webauthn-protocol/) explains the many pros of webauthn as well as the few cons.

tests/urls.py
```diff
@@ -1,4 +1,5 @@
 from django.conf.urls import include, url
+from django.contrib import admin
```
moggers87, line 2: Is this import necessary?

two_factor/views/core.py
```diff
@@ -199,6 +246,43 @@ def get_context_data(self, form, **kwargs):
...
         return context
 
+    def get_remember_agent(self):
+        """
+        Returns if a user, browser and device is remembered using the remember cookie.
+        Uses a cached value if called before.
+        """
+        if self.remember_agent_cache is None:
+            self.remember_agent_cache = self._remember_agent()
+        return self.remember_agent_cache
```
moggers87, line 256: Django has a `cached_property` decorator which we should use rather than doing it ourselves: https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.functional.cached_property

two_factor/views/utils.py
```diff
@@ -194,3 +197,57 @@ def simple_decorator(View):
...
+
+def get_remember_device_cookie(user_pk, password_hash, otp_device_id):
+    """
+    Compile a signed cookie from user_pk, password_hash and otp_device_id,
+    but only return the hashed and signatures and omit the data.
+
+    The cookie is composed of 3 parts:
+    1. A weakly hashed value of otp_device_id.
+    2. A timestamp of signing.
+    3. A hmac signature that signs user_pk, password_hash, otp_device_id and the timestamp.
```
moggers87, line 210: I'm not keen on using the password hash as this data will be left in the user's browser. Is there a particular reason it's being used?

two_factor/forms.py
```diff
@@ -169,6 +170,14 @@ def __init__(self, user, initial_device, **kwargs):
                 isinstance(initial_device, (RemoteYubikeyDevice, YubikeyDevice)):
             self.fields['otp_token'] = forms.CharField(label=_('YubiKey'), widget=forms.PasswordInput())
 
+        # Add a field to remeber this browser.
+        if getattr(settings, 'TWO_FACTOR_REMEMBER_COOKIE_AGE', None):
+            self.fields['remember'] = forms.BooleanField(
+                required=False,
+                initial=True,
+                label=_("Don't ask again on this device")
```
moggers87, line 178: I think the user needs to be informed how long their second factor will be remembered for. Especially important on the example site where it's only going to be remembered for 60 seconds :laughing:

moggers87:
@beckedorf @compsult I think webauthn should be discussed in a separate issue

two_factor/views/utils.py
beckedorf: I would prefer a solution without using the password hash as well. I included it to archive that all remembered devices become invalid once a the user changed his password. I think this is a good safety measure, as users (hopefully) reset their passwords after a breach but probably don't care about remembered devices.

I had a bad feeling as well storing the password hash in the user's browser. This is why I am only storing the signature and omit the data. A couple of lines below the data is cleaned from the cookie

two_factor/forms.py
beckedorf: I agree!

two_factor/views/utils.py
beckedorf: Everything but this question is fixed. 😄

moggers87: The signature still depends on the password hash, so there's always a chance that there could be some information leakage.

How about generating a random number that is stored in the cache for each user? It could then be deleted/regenerated when the user changes their password. You'd have to check this via the `pre_save` signal, something like:

```
@receiver(pre_save, sender=get_user_model())                                                                                                                                                                                                   
def user_updated_password(sender, **kwargs):                                                                                                                                                                                                   
    user = kwargs.get('instance', None)                                                                                                                                                                                                        
    new_pass = user.password                                                                                                                                                                                                                   
    try:                                                                                                                                                                                                                                       
        old_pass = get_user_model().objects.get(pk=user.pk).password                                                                                                                                                                           
    except get_user_model().DoesNotExist:                                                                                                                                                                                                      
        old_pass = None                                                                                                                                                                                                                        
    if old_pass != new_pass:                                                                                                                                                                                                                   
        # invalidate cache    
```

beckedorf: I agree. Better be save than sorry.

What I like about this solution is far, is that the persistent store does not need to be changed. I guess for a caching solution we would have to alter the model of some sort?

Another approach would be to use only the salt of the encrypted password as part of encrypted data. What do you think?

beckedorf: I just modified the code accordingly. Looking forward to your opinion.

beckedorf: I just stumbled upon the PasswordResetTokenGenerator which also uses the password hash to create as hash sum and send this value to the user.

https://github.com/django/django/blob/da4923ea87124102aae4455e947ce24599c0365b/django/contrib/auth/tokens.py#L91

So I guess I theory it should be save to use hash the password hash. Just to be save we could implement it a bit more similar to the PasswordResetTokenGenerator. What do you think?

moggers87: With regards to using the (plaintext!) salt rather than the password hash: that's a huge no-no.

I still don't feel super comfortable with using the password hash either, but `TimestampSigner` is using all the same things that `PasswordResetTokenGenerator` does so :woman_shrugging:

So: revert your last two commits and we'll carry on from there :smile_cat:

beckedorf: I went back to the old technique but removed the dependancy to `TimestampSigner`. That class was not a good fit as we do not want to store the whole signed value in a cookie. Instead I used PasswordResetTokenGenerator as a basis.

In the next version Django will move from sha1 to sha256. To increase security and prevent legacy problems I already rely on the new algorithm.

Feel free to take it from here... :-D

Bouke:
I really like this functionality, how can we take this further?

moggers87:
It just needs another review. I guess this one fell between the cracks in my TODO list :laughing:

beckedorf:
From my point of view it is ready forge merge. But since the changes are more than minor it would be good if @Bouke , @moggers87 or someone else with a good overview of the structure of this package double check that it fits into the larger project.

two_factor/views/core.py
```diff
@@ -218,6 +266,44 @@ def get_context_data(self, form, **kwargs):
...
+    @cached_property
+    def remember_agent(self):
+        """
+        Returns True if a user, browser and device is remembered using the remember cookie.
+        """
+        if not getattr(settings, 'TWO_FACTOR_REMEMBER_COOKIE_AGE', None):
+            return False
+
+        user = self.get_user()
+        if user:
```
moggers87, line 278: This line has no coverage. However it's not due to a lack of tests, but rather it's only ever used in conjunction with calls to `default_device` which already checks against user being `None`.

I'm fairly sure this line can be removed and the rest of the function will work fine.

two_factor/views/utils.py
```diff
@@ -194,3 +202,72 @@ def simple_decorator(View):
...
+
+def hash_remember_device_cookie_value(otp_device_id, user, timestamp):
+    salt = 'two_factor.views.utils.hash_remember_device_cookie_value'
+    value = otp_device_id + str(user.pk) + str(user.password) + timestamp
+    return salted_hmac_sha256(salt, value).hexdigest()
+
+
+# inspired by django.utils.crypto.salted_hmac django versions > 3.1a1
+def salted_hmac_sha256(key_salt, value, secret=None):
+    if secret is None:
```
moggers87, line 268: No coverage for this line, it just needs a simple test to make sure `salted_hmac_sha256("blah", "blah")` returns a different value from `salted_hmac_sha256("blah", "blah", "my-new-secret")`

docs/configuration.rst
```diff
@@ -125,3 +125,73 @@ Fake Gateway
...
+
+During a successful login with a token, the user may choose to remember this browser.
+If the same user logs in again on the same browser, a token will not be requested, as the browser
+serves as a second factor.
+
+The option to remember a browser is deactived by default. Set `TWO_FACTOR_REMEMBER_COOKIE_AGE` to activate.
+
+The browser will be remembered as long as
+- the cookie that authorizes the browser did not expire,
+- the user did not reset his password, and
```
moggers87, line 140: Custom users might not use `password` to store their password, .e.g [django-auth-ldap](https://django-auth-ldap.readthedocs.io/en/latest/users.html#user-objects)

As far as I can see, the code will work just fine but we probably should warn users if they use a custom user then remember tokens might not get reset.

docs/configuration.rst
```diff
@@ -125,3 +125,73 @@ Fake Gateway
...
+Remember Browser
+----------------
+
+During a successful login with a token, the user may choose to remember this browser.
+If the same user logs in again on the same browser, a token will not be requested, as the browser
+serves as a second factor.
+
+The option to remember a browser is deactived by default. Set `TWO_FACTOR_REMEMBER_COOKIE_AGE` to activate.
+
+The browser will be remembered as long as
```
moggers87, line 138: There needs to be a new line between this and the list below, otherwise the list doesn't render correctly

docs/configuration.rst
moggers87: Also that should be "their" not "his"

moggers87:
:tada:

beckedorf:
🚀

PR #362, opened 2020-07-04
Push builds to pypi

Bouke:
This workflow pushes builds from the master branch automatically to [TestPyPI](https://test.pypi.org/project/django-two-factor-auth/#history) and tagged commits to normal PyPI automatically. So the workflow would be something like this:

1. Pull translations (we will probably integrate Transifex to make this automatic)
2. Tag the commit with the version number (e.g. `1.11.1`) and push the tag. Alternatively create a release in GitHub, which will also create a tag.
3. Workflow is triggered: will create the build and push to pypi

I've been testing the workflow on TestPyPI, but not to real PyPI yet. There is a small, yet deliberate, change in version numbering. Bumpversion will now leave out the patch if zero. So 1.12 instead of 1.12.0. PEP440 says that those versions are the same number, so I expect no issues from this.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,19 @@
 [bumpversion]
 current_version = 1.11.0
-files = setup.py docs/conf.py
 commit = True
 tag = True
 tag_name = {new_version}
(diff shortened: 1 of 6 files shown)


Bouke:
Interestingly isort v5 just got released an hour ago and breaks the build. So I've included a fix for that as well.

PR #368, opened 2020-07-25
Minor cleanups

moggers87:
## Description

- Remove references to Django 1.11
- Remove dead code
- Update example code in docs
- Test should print warnings
- Replace deprecated functions
  * Replace `url()` with `re_path()`
  * Replace assertEquals with assertEqual
- Resolve LOGOUT_URL warning (and add tests for it)
- Run flake8 on the same files as isort
  * And fix flake8 errors that weren't being reported before

## Motivation and Context

Improves coverage, future proofs code, and declutters tox.ini

## How Has This Been Tested?

CI passes

## Types of changes

## Checklist:

### DIFF
--- a/example/urls.py
+++ b/example/urls.py
@@ -1,7 +1,7 @@
 from django.conf import settings
-from django.conf.urls import include, url
 from django.contrib import admin
 from django.contrib.auth.views import LogoutView
+from django.urls import include, re_path
 
 from two_factor.gateways.twilio.urls import urlpatterns as tf_twilio_urls
 from two_factor.urls import urlpatterns as tf_urls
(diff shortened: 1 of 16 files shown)


claudep:
Note you might also consider replacing the `url()` functions by `path()` in  `urls.py`.

moggers87:
Did I miss anything?

example/urls.py
```diff
@@ -11,39 +11,39 @@
...
         name='registration_complete',
     ),
-    url(r'', include(tf_urls)),
-    url(r'', include(tf_twilio_urls)),
-    url(r'', include('user_sessions.urls', 'user_sessions')),
-    url(r'^admin/', admin.site.urls),
+    re_path(r'', include(tf_urls)),
+    re_path(r'', include(tf_twilio_urls)),
+    re_path(r'', include('user_sessions.urls', 'user_sessions')),
+    re_path(r'^admin/', admin.site.urls),
```
claudep, line 42: `re_path` should be limited to cases where a regular expression is unavoidable. In this file, I don't see any of these, so using `path` is IMHO better and simpler (e.g. `path('account/register/',…)`).

tests/test_views_login.py
```diff
@@ -450,29 +450,46 @@ def test_generate(self):
                          response.context_data['device'].token_set.all()])
         self.assertNotEqual(first_set, second_set)
 
+    def test_no_cancel_url(self):
+        response = self.client.get(reverse('two_factor:login'))
+        self.assertEqual(response.status_code, 200)
+        self.assertNotIn('canel_url', response.context.keys())
```
claudep, line 456: `cancel_url`?

tests/urls.py
```diff
@@ -8,44 +8,44 @@
...
-                                verification_url='/account/login/'),
+    re_path(
+        r'^secure/redirect_unverified/$',
+        SecureView.as_view(raise_anonymous=True,
+                           verification_url='/account/login/'),
     ),
-    url(r'', include(tf_urls)),
-    url(r'', include(tf_twilio_urls)),
+    re_path(r'', include(tf_urls)),
+    re_path(r'', include(tf_twilio_urls)),
```
claudep, line 50: Same remark than for main `urls.py`, `path` should be privileged.

two_factor/gateways/twilio/urls.py
```diff
@@ -1,11 +1,11 @@
...
+from django.urls import re_path
 
 from .views import TwilioCallApp
 
 urlpatterns = ([
-    url(
-        regex=r'^twilio/inbound/two_factor/(?P<token>\d+)/$',
-        view=TwilioCallApp.as_view(),
+    re_path(
+        r'^twilio/inbound/two_factor/(?P<token>\d+)/$',
```
claudep, line 7: Even here, we could use `path('twilio/inbound/two_factor/<int:token>/',…)`

example/urls.py
moggers87: Using `re_path` means we're swapping like for like (`url` just calls `re_path`). I think we should leave migrating to `path` to another PR.

claudep: Sure, that makes sense.

claudep: Done in #374.

PR #369, opened 2020-07-27
Add redirect_field_name to LoginView context to fix "next"

merito:
## Description
Based on https://github.com/django/django/blob/f4e93919e4608cfc50849a1f764fd856e0917401/django/contrib/auth/views.py#L99

## Motivation and Context
It resolves https://github.com/Bouke/django-two-factor-auth/issues/214

## How Has This Been Tested?
Tested with Django==3.0.7 and it works as expected.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/views/core.py
+++ b/two_factor/views/core.py
@@ -334,6 +334,7 @@ class LoginView(SuccessURLAllowedHostsMixin, IdempotentSessionWizardView):
                 "review the URL and update your settings.",
                 DeprecationWarning)
             context['cancel_url'] = resolve_url(settings.LOGOUT_URL)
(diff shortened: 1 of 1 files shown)


merito:
> There need to be tests. In particular, there need to be tests that would fail without this fix.

True. I've tried to write a test which fails without this change and I've realized that the issue is not there ;)

ex5:
I've just ran into the same issue: our existing service relies on `next` being handled the way it's handled in the builtin `LoginView`, and this exact change fixes the problem.

> I've tried to write a test which fails without this change and I've realized that the issue is not there ;)

@merito could you elaborate on this? What do you mean "the issue is not there"?

Otherwise I'd like to try and make the missing test if it helps getting this merged.

PR #371, opened 2020-08-10
Add autofocus, autocomplete one-time-code and inputmode numeric to token input fields

frjo:
## Description

By adding some extra attributes to the token input field the users gets a wider experience.

* Autofocus makes sure the users starts typing in the correct inout field. This was already added to the login token field but not when setting 2FA up for the first time.
* Autocomplete one-time-code is supported in Safari (maybe others as well?) and allow the browser to suggest codes send via SMS etc.
* Inputmode numeric makes surety best keyboard are selected on mobil.

## Motivation and Context

* https://www.twilio.com/blog/html-attributes-two-factor-authentication-autocomplete
* https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Values

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] My code follows the code style of this project.

### DIFF
--- a/two_factor/forms.py
+++ b/two_factor/forms.py
@@ -58,6 +58,10 @@ class PhoneNumberForm(ModelForm):
 class DeviceValidationForm(forms.Form):
     token = forms.IntegerField(label=_("Token"), min_value=1, max_value=int('9' * totp_digits()))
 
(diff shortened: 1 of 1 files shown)


moggers87:
Closing favour of #396

PR #374, opened 2020-08-13
Simplified re_path() to path() URLs

claudep:


two_factor/gateways/twilio/views.py
```diff
@@ -80,5 +80,5 @@ def get_prompt_context(self):
             # Build the prompt. The numbers have to be clearly pronounced,
             # this is by creating a string like "1. 2. 3. 4. 5. 6.", this way
             # Twilio reads the numbers one by one.
-            'token': '. '.join(self.kwargs['token']) if self.request.method == 'POST' else '',
+            'token': '. '.join(str(self.kwargs['token'])) if self.request.method == 'POST' else '',
```
moggers87, line 83: I see the broken build on travis, but I'm not sure this is breaking because of the urlconf changes.

I've started another build of master to check

moggers87: Build on master passes, so why does this break when `path` is used? :thinking:

claudep: Because with `<int:variable>` path syntax, the returned value has been converted to an `int`, contrary to the regex syntax which always keep string values.

PR #375, opened 2020-08-14
Removed upper limit of django-otp dependency

claudep:
django-otp recently released a 1.0.0 version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
         'Django>=2.2',
-        'django_otp>=0.6.0,<0.99',
+        'django_otp>=0.6.0',
         'qrcode>=4.0.0,<6.99',
         'django-phonenumber-field>=1.1.0,<3.99',
         'django-formtools',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -13,7 +13,7 @@
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
         'Django>=2.2',
-        'django_otp>=0.6.0,<0.99',
+        'django_otp>=0.6.0',
```
moggers87, line 16: Do you think it's safe to have no upper limit? django-otp does publish quite often and has been known to break things at random. Will that change post 1.0.0?

claudep: I think that tests should show when some `django_otp` release breaks things. That can happen with any other package, I don't really see a good reason to special-case `django_otp` at this regard.

moggers87:
Oops, that should have been approved, not request for changes.

moggers87:
@claudep ???

claudep:
Sorry, I thought this was merged already! I guess I have to recreate a new PR.

moggers87:
There should be a button to restore the branch.

claudep:
Ah yes, thanks!!

moggers87:
Thanks! :+1:

PR #376, opened 2020-08-14
Added testing for Django 3.1

claudep:


tox.ini
```diff
@@ -51,6 +54,7 @@ deps =
     dj111: Django<1.12
```
moggers87, line 54: There are still some references to Django 1.11 in `tox.ini`. This is a good time to remove those too.

PR #379, opened 2020-08-20
Fixed #364 - Replaced Bootstrap 3 by Bootstrap 4.

### LINKED ISSUE #364: Upgrade templates to Bootstrap 4
Several buttons in this package templates are still using the `btn-default` style which is gone in Bootstrap 4. This means that those buttons are not decorated at all and don't look like buttons with Boostrap 4. The `btn-default` style could be kept for backwards compatibility, but a new class should be added on those buttons, maybe `btn-outline-secondary` (as there are no direct equivalents of `btn-default` in Boostrap 4 button styles).

claudep:
I did my best to migrate Bootstrap classes. Might be worth double-checking the result.

### DIFF
--- a/example/templates/_base.html
+++ b/example/templates/_base.html
@@ -4,35 +4,33 @@
 <head>
   <title>{% block title %}{% endblock %}</title>
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <link href="//cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.0.2/css/bootstrap.min.css" rel="stylesheet" media="screen">
+  <link href="//cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.5.2/css/bootstrap.min.css" rel="stylesheet" media="screen">
+  <script defer src="//cdnjs.cloudflare.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+  <script defer src="//cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.5.2/js/bootstrap.min.js"></script>
 </head>
 <body>
   {% block content_wrapper %}
-    <nav class="navbar navbar-default" role="navigation">
+    <nav class="navbar navbar-light" role="navigation">
       <div class="container">
         <!-- Brand and toggle get grouped for better mobile display -->
-        <div class="navbar-header">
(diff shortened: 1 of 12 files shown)


example/templates/_base.html
```diff
@@ -4,35 +4,33 @@
...
                 <a href="{% url 'home' %}">{% trans "Home" %}</a></li>
             <li class="{% block nav_secret %}{% endblock %}">
                 <a href="{% url 'secret' %}">{% trans "Secret Page" %}</a></li>
           </ul>
-          <ul class="nav navbar-nav navbar-right">
+          <ul class="navbar-nav ml-auto">
             {% if user.is_authenticated %}
               <li><a>{{ user }}
-                {% if user.otp_device %}<span class="glyphicon glyphicon-lock"></span>{% endif %}
+                {% if user.otp_device %}<i class="fa fa-lock"></i>{% endif %}
```
moggers87, line 33: Is Font Awesome bundled with Bootstrap 4? Does Font Awesome provide a CDN for us to use?

Alternatively: there's a lock icon in Unicode that could be used instead.

claudep: I considered the lock icon Unicode, but it looks like not very widely supported by common fonts. Did you experiment it?

And no, Font Awesome is not bundled.

claudep: I replaced the Font Awesome tag by the Unicode version of the character.

mathiasertl:
@claudep I have run the merged PR and think it's fine. LGTM!

moggers87 (changes requested):
Hey it's only been a month since you opened this PR :tada: 

Looks fine for the most part, there's only one thing that bugs me:

Bootstrap 3 navbar:
Boostrap 3 with a gap under the nav

Bootstrap 4 navbar, notice the lack of spacing between it and the page content:
Boostrap 4 with no gap under the nav

I don't know whether this is the new default or if there needs to be change in the markup to give the navbar some margin underneath, but it really sticks out to me as if it was a mistake either way.

claudep:
Indeed, `.navbar` has a margin bottom of 20px on BS3, and none on BS4. If you'd like to keep that spacing, we can simply add a `mb-3` style to the navbar to get a similar result.

moggers87:
Much better. I'm chalking up any remaining "it looks weird" to me not being used to Boostrap 4 :stuck_out_tongue_winking_eye:

moggers87 (approved):

PR #380, opened 2020-08-25
Add the possibility for the user to change the double authentication

lmarion-source:
Add the possibility for the user to change the double authentication

## Description

From the profile view, the user can now switch between Google authenticator/ sms or call method for double authentication.

## Motivation and Context
This solves the issue reported in https://github.com/Bouke/django-two-factor-auth/issues/347

## How Has This Been Tested?
No new test was added, however the example repository has been updated to take the change into account. This can then be tested locally by using the example repository

## Types of changes
- [ X ] New feature (non-breaking change which adds functionality)

## Checklist:
- [ X ] My code follows the code style of this project.
- [ X ] All existing tests passed.


two_factor/views/core.py
```diff
@@ -530,9 +530,12 @@ def get_device(self, **kwargs):
 
     def get_key(self, step):
         self.storage.extra_data.setdefault('keys', {})
+
         if step in self.storage.extra_data['keys']:
             return self.storage.extra_data['keys'].get(step)
+
         key = random_hex_str(20)
+
```
moggers87, line 538: Why all these random newlines? I don't see how they are related to your proposed change.

two_factor/views/core.py
```diff
@@ -700,6 +705,99 @@ def get_context_data(self):
...
+    """
+    View for changing the two-factor authentication method from phone number to token generator.
+    """
+    template_name = 'two_factor/core/setup_reset_generator.html'
+    success_url = settings.LOGIN_REDIRECT_URL
+    qrcode_url = 'two_factor:qr'
+    session_key_name = 'django_two_factor-qr_secret_key'
+
+    form_list = (
+        ('generator', TOTPDeviceForm),
```
moggers87, line 720: There's only one form here, what about yubikey and phone methods?

moggers87 (changes requested):
That's quite a lot of templates you've added to the example app and I'm not quite sure why they're necessary (and why those changes aren't just in `two_factor` itself if they are necessary).

Also tests are not optional!

lmarion-source:
I added a bunch of templates because the example was not working initially....a bunch of templates were actually missing to make it work (or I missed something?). I'll write tests.

two_factor/views/core.py
lmarion-source: This is exactly the point: this view is used when you already have a phone as double authentication method and want to go to the generator.

moggers87: What if I want to change to another method?

lmarion-source: Well then you have the other view that allows you to go from the token generator to the SMS/call. You can only have one method for double authentication at a time. If you previously were using the google authenticator, you'll arrive on the view to use SMS/call. If you previously were using the SMS/call, you'll arrive on the view to use google authenticator. If you want to add extra backup tokens or phone etc nothing changes. All of this assumes you already have enable the double authentication method. If not, then you are still pointing to the initial setup view.

moggers87: And what about yubikey?

lmarion-source: I will add this one. My bad.

lmarion-source: tests are added, the case of yhe use of a yubikey is also considered

PR #383, opened 2020-08-29
Ability to send custom sms using Twilio during OTP authentication

aseem-hegshetye:
Currently the default sms thats sent by using Twilio during 2FA is -
`Your authentication token is %s`

With this PR we can customize the message by setting variable `settings.TWILIO_MESSAGE`
Example:
`settings.TWILIO_MESSAGE='Your authentication token is {}'`


tests/test_gateways.py
```diff
@@ -59,8 +61,12 @@ def test_gateway(self, client):
                 url='http://testserver/twilio/inbound/two_factor/%s/?locale=en-us' % code)
 
             twilio.send_sms(device=Mock(number=PhoneNumber.from_string('+123')), token=code)
+
             client.return_value.messages.create.assert_called_with(
-                to='+123', body='Your authentication token is %s' % code, from_='+456')
+                to='+123',
+                body=getattr(settings, 'TWILIO_MESSAGE').format(code),
+                from_='+456'
+            )
```
moggers87, line 69: Would be nice to have a test with the default string too

two_factor/gateways/twilio/gateway.py
```diff
@@ -52,7 +53,14 @@ def make_call(self, device, token):
...
 
     def send_sms(self, device, token):
-        body = gettext('Your authentication token is %s') % token
+        """
+        send sms using custom TWILIO_MESSAGE from settings.py
+        """
+        body = getattr(
+            settings,
+            'TWILIO_MESSAGE',
+            'Your authentication token is {}'
```
moggers87, line 62: I think the default value should still be a translated string. For most people this has been fine and they might be a little annoyed when their users start getting English text messages

two_factor/gateways/twilio/gateway.py
```diff
@@ -52,7 +53,14 @@ def make_call(self, device, token):
                                  url=uri, method='GET', timeout=15)
 
     def send_sms(self, device, token):
-        body = gettext('Your authentication token is %s') % token
+        """
+        send sms using custom TWILIO_MESSAGE from settings.py
+        """
+        body = getattr(
+            settings,
+            'TWILIO_MESSAGE',
```
moggers87, line 61: I think a better name for this setting would be `TWO_FACTOR_SMS_MESSAGE`. It should also be documented so users know they can change it.

moggers87 (changes requested):
Overall this looks good :+1:

There's just a few minor changes I'd like to see before merging.

claudep:
An alternative would be to use a template, to not add yet another setting. Less straightforward to configure than a setting, but a lot more powerful as you can pass more context information.

moggers87:
@claudep yet another template though :stuck_out_tongue: but I agree it's a better solution.

aseem-hegshetye:
I have updated the PR with everything requested. Please let me know if more changes are required

two_factor/gateways/twilio/gateway.py
```diff
@@ -51,8 +53,15 @@ def make_call(self, device, token):
                                  from_=getattr(settings, 'TWILIO_CALLER_ID'),
                                  url=uri, method='GET', timeout=15)
 
-    def send_sms(self, device, token):
-        body = gettext('Your authentication token is %s') % token
+    def send_sms(self, device, token, sms_message):
```
moggers87, line 56: This method shouldn't require any extra parameters.

two_factor/templates/two_factor/twilio/sms_message.html
```diff
@@ -0,0 +1,3 @@
+{% load i18n %}
+{% trans sms_message %}
```
moggers87, line 2: This doesn't need to be a variable, if users want to override the default message they will override the template itself.

two_factor/templates/two_factor/twilio/sms_message.html
```diff
@@ -0,0 +1,3 @@
...
+{% trans 'Your token is '%}{{ token }}
```
moggers87, line 3: This should use `blocktrans` so translators have some context. Also the file should end with a newline.

moggers87 (changes requested):
I'm not actually sure why tests are passing (unless I'm missing something, your changes will break sending SMS). That's not something that should be fixed in this PR, but it's something to be aware of :cat:

two_factor/templates/two_factor/twilio/sms_message.html
aseem-hegshetye: done

two_factor/gateways/twilio/gateway.py
aseem-hegshetye: done

two_factor/templates/two_factor/twilio/sms_message.html
aseem-hegshetye: done

aseem-hegshetye:
Anything else I need to do for this PR?

moggers87:
I need to re-review it and either merge or ask for more changes

tests/test_gateways.py
```diff
@@ -98,6 +109,35 @@ def test_invalid_twilio_language(self, client):
             with translation.override('ar'):
                 twilio.make_call(device=Mock(number='+123'), token='654321')
 
+    @override_settings(
+        TWILIO_ACCOUNT_SID='SID',
+        TWILIO_AUTH_TOKEN='TOKEN',
+        TWILIO_CALLER_ID='+456',
+    )
+    @patch('two_factor.gateways.twilio.gateway.Client')
+    def test_twilio_blank_sms_message(self, client):
```
moggers87, line 118: I think we can remove this test as we're now using a templates.

two_factor/templates/two_factor/twilio/sms_message.html
```diff
@@ -0,0 +1,5 @@
+{% load i18n %}
+{% blocktrans %}
```
moggers87, line 2: `{% blocktrans trimmed %}` is better for translators (it removes unnecessary whitespace)

moggers87 (changes requested):
Two more minor tweaks and then I think this is ready to merge :+1:

aseem-hegshetye:
I have made all requested changes

moggers87 (approved):

moggers87:
@aseem-hegshetye thanks for working on this :smiley_cat:

PR #386, opened 2020-09-17
Hide submit button using class instead of left-margin, makes it usable with CSP

mathiasertl:
Hide submit button using class instead of left-margin, makes it usable with CSP

## Description

Replace inline CSS styles with existing bootstrap CSS class to hide the submit button.

I'm still a bit unhappy with the empty paragraph as hack. Without it, the table and the submit button row has no space between them. I can think of two solutions:

* Upgrade to bootstrap4 (which has appropriate margin classes) and add them to the table.
* Add a static CSS file with a custom class that adds that same margin.

If any of the two seem better to you, please let me know, I'm happy to amend the PR.

## Motivation and Context

This makes the default templates usable with a restrictive [Content-Security Policy](https://en.wikipedia.org/wiki/Content_Security_Policy).

## How Has This Been Tested?

I could not find any unittests for the UI. If there are, please let me know.

I have manually tested the views, and also tried the enter button of course (that this submit button seems to facilitate).

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/templates/two_factor/_wizard_forms.html
+++ b/two_factor/templates/two_factor/_wizard_forms.html
@@ -1,6 +1,4 @@
-<table>
+<table class="mb-3">
   {{ wizard.management_form }}
   {{ wizard.form }}
 </table>
-{# Add empty paragraph to add some space below the table #}
-<p></p>
--- a/two_factor/templates/two_factor/_wizard_forms.html
+++ b/two_factor/templates/two_factor/_wizard_forms.html
@@ -2,3 +2,5 @@
   {{ wizard.management_form }}
   {{ wizard.form }}
 </table>
+{# Add empty paragraph to add some space below the table #}
+<p></p>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - The templates are now based on Bootstrap 4.
 - `DisableView` now checks user has verified before disabling two-factor on
   their account
+- Inline CSS has been removed  to allow stricter Content Security Policies.
 
 ## 1.12 - 2020-07-08
 ### Added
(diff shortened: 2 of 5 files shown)


claudep:
Note that the patch to move to Bootstrap 4 (#379) is only waiting for a final review.

mathiasertl:
Aha! Excellent. Then I propose we wait wait for that PR and I will adapt mine then. 

In the meantime anything else I can do to improve the PR, or is it otherwise fine?

moggers87:
> Note that the patch to move to Bootstrap 4 (#379) is only waiting for a final review.

noted

two_factor/templates/two_factor/_wizard_forms.html
```diff
@@ -2,3 +2,5 @@
   {{ wizard.management_form }}
   {{ wizard.form }}
 </table>
+{# Add empty paragraph to add some space below the table #}
+<p></p>
```
moggers87, line 6: If there needs to be extra space, then you need to add margin to the table. I think Boostrap has a class for that.

moggers87 (changes requested):
If you could update your branch with the latest commits from master as well please.

two_factor/templates/two_factor/_wizard_forms.html
mathiasertl: Bootstrap 4 has, Bootstrap 3 didn't ;-), that's why I added the paragraph-hack. Now updating the branch.

mathiasertl:
I have updated the branch to use bs4 classes, and updated the changelog as well.

@claudep is there a reason the example project overrides some of the templates from the app? They seem like unnecessary code duplication to me. If you agree, I can make a separate PR for that.

claudep:
I'm not the author of the example files (and don't care much about them, frankly).

mathiasertl:
Alright. Then I propose we merge this as is.

moggers87 (approved):

PR #390, opened 2020-10-01
DisableView should require request user to be verified

### LINKED ISSUE #388: DisableView doesn't require request user to be verified
`DisableView` only checks that the request user *has* a valid, not that the user is verified.

It is possible that if a site has been misconfigured and allows users to bypass django-two-factor-auth's login view (e.g. the admin site hasn't been configured correctly and admins can login without 2fa) and then disable that user's two-factor device(s). At this point an attacker could go set up a new two-factor and gain access to views protected by `otp_required`.

This bug could also be exploited if the user has multiple sessions open before enabling two-factor auth. Any other session could then disable two-factor auth without having access to the device.

## Expected Behavior

`DisableView` should check if the current user has two factor auth enabled on their account *and* has used it to log in on this session.

## Current Behavior

`DisableView` only checks that the user has two factor auth enabled on their account.

## Possible Solution

Swap `login_required` for `otp_required`. Write some tests too.

```
>>> from two_factor.views.profile import DisableView                                                                                                                                                   
>>> from django.contrib.auth import get_user_model
>>> from django.http import HttpRequest
>>> user = get_user_model().objects.first()
>>> user.staticdevice_set.get_or_create(name='backup')
>>> print(getattr(user, "is_verified", "user has no is_verified method"))
user has no is_verified method
>>> request = HttpRequest()
>>> request.user = user
>>> request.method = "GET"
>>> DisableView.as_view()(request)
<TemplateResponse status_code=200, "text/html; charset=utf-8">
>>> request.method = "POST"
>>> request.POST = {"understand": "on"}
>>> DisableView.as_view()(request)
<HttpResponseRedirect status_code=302, "text/html; charset=utf-8", url="/user/home/">
>>> print(user.staticdevice_set.count())
0
```

It should also be possible to reproduce this bug by logging in from two different browsers, enabling two factor auth on one session and then disabling it from the other (I've not tried it due to time constraints) 

## Context

<https://github.com/Bouke/django-two-factor-auth/issues/253#issuecomment-701881736> prompted me to look at the source code.

## Your Environment

* Browser and version: n/a, this was done via the shell
* Python version: 3.6.12
* Django version: 2.2.15
* django-otp version: 0.9.4
* django-two-factor-auth version: 1.12.1
* Link to your project: https://github.com/Bouke/django-two-factor-auth

moggers87:
## Description

Replaced `login_required` with `otp_required`

## Motivation and Context

Fixes #388

## How Has This Been Tested?

Tests written, DisableView tests have also been refactored slightly too

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Changed
 - The templates are now based on Bootstrap 4.
+- `DisableView` now checks user has verified before disabling two-factor on
+  their account
 
 ## 1.12 - 2020-07-08
 ### Added
--- a/tests/test_views_disable.py
+++ b/tests/test_views_disable.py
@@ -2,7 +2,7 @@ from django.conf import settings
 from django.shortcuts import resolve_url
 from django.test import TestCase
 from django.urls import reverse
-from django_otp import devices_for_user
+from django_otp import DEVICE_ID_SESSION_KEY, devices_for_user
 
 from .utils import UserMixin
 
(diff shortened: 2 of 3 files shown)


schinckel:
IMHO Backup token view should also require verification.

(Actually, backup tokens should not be stored as plain-text, but it's still the same problem - an unverified but authenticated user can bypass 2FA by viewing or regenerating tokens, and then using one of these).

moggers87:
The backup token view is already protected: https://github.com/Bouke/django-two-factor-auth/blob/4031ca209aca0127f0cd3ddedc1b94526a5ca78d/two_factor/views/core.py#L566

schinckel:
Oh, I guess I never actually clicked on the backup view.

What about the profile view itself? Is there any reason a non-verified user should be viewing that, I wonder?

moggers87:
What's on there that needs protecting?

schinckel:
Probably nothing: but lots of sites will require you to verify before you even get to pages that list are security-related (even if they don't themselves contain any sensitive data).

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Changed
 - The templates are now based on Bootstrap 4.
+- `DisableView` now properly checks user has verified before disabling
+  two-factor on their account
 
 ## 1.12 - 2020-07-08
```
Bouke, line 8: I’d remove “properly” here. We don’t support any other login flows than our own, thus 2fa was already verified on sign-in. From the docs:

> Warning:  
> Be sure to remove any other login routes, otherwise the two-factor authentication might be circumvented.

Bouke (approved):

CHANGELOG.md
moggers87: Done :+1:

PR #391, opened 2020-10-03
Removed obsolete providing_args Signal argument.

claudep:
The purely documentational providing_args argument was deprecated
in Django 3.1.


moggers87 (approved):
:+1:

PR #395, opened 2020-10-20
Unify the text of the "back to profile" links

sgaist:
## Description

The target page of the "two_factor:profile" path is in fact Account Security 
and using "Back to Profile" is a bit misleading as users may
expect to be back to their own profile page. Therefore
use the same text as the backup_tokens template to be
more consistent.

## Motivation and Context

The two links bring back to the same "Account Security" page but the text 
mentioning "Profile" rather suggests that the user profile page will be shown 
which is not what will happen. 

Note that there is no need to for new translation as the text itself already exists.

## How Has This Been Tested?

Simply use the package as usual, it is just a text change.

## Types of changes

- [x] Template text change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - `DisableView` now checks user has verified before disabling two-factor on
   their account
 - Inline CSS has been removed  to allow stricter Content Security Policies.
+- Change "Back to Profile" to "Back to Account Security"
 
 ## 1.12 - 2020-07-08
 ### Added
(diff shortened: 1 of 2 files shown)


sgaist:
For the translation part, I do not know what would be simpler. There's no new text to translate, it's rather moving some lines around the translation files.

moggers87:
If it's is the exact same string, then it should work just fine. I think the line numbers in the PO files are just there as hints.

moggers87:
I'll do a review once Travis tests pass

moggers87 (approved):
Looks  
Good  
To  
Me

Issue #387, opened 2020-09-21
Please update the django-phonenumber-field dependency to allow version 5.0.0


Lionqueen94:
Please update the django-phonenumber-field dependency to allow version 5.0.0.
Currently only up to version 3.99 is allowed, and in the meantime version 4.0.0 and 5.0.0 came out in the past year.

moggers87:
Are there any changes in those versions that might effect django-two-factor-auth users?

syserr0r:
Not sure how your latest release can claim 'Django 3.1 support' when your dependencies do not.

Django 3.1 support was only added to django-phonenumber-field in 5.0 (see https://github.com/stefanfoulis/django-phonenumber-field/blob/master/CHANGELOG.rst)

moggers87:
That changelog also claims that support for Django 2.2 has been dropped. Is that a mistake?

moggers87:
Apparently yes, it is a mistake https://github.com/stefanfoulis/django-phonenumber-field/issues/382#issuecomment-684765218

atkawa7:
ping @moggers87

moggers87:
Closed by #401

PR #401, opened 2020-12-21
Update django-phonenumber-field dependency to newer version

syserr0r:
Untested attempt to allow newer django-phonenumber-field dependency for issue #387

## Description
As above
PS: I would like to know why you required <3.99 and not <4 -- is this best practice or for some other reason?

## Motivation and Context
Issue #387

## How Has This Been Tested?
It hasn't.
https://github.com/stefanfoulis/django-phonenumber-field/blob/master/CHANGELOG.rst reports no breaking changes so fingers crossed, eh?

N/a

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
  - Provide Django 3.1 support in dependency
  - Allow package use when people are already using django-phonenumber-field 4 or newer

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         'Django>=2.2',
         'django_otp>=0.8.0',
         'qrcode>=4.0.0,<6.99',
-        'django-phonenumber-field>=1.1.0,<3.99',
+        'django-phonenumber-field>=1.1.0,<6',
         'django-formtools',
     ],
     extras_require={


moggers87:
No, "fingers crossed" is not good enough. Actual user testing is required, i.e. install 5.0.0 and see if you can still set a phone number.

claudep:
Shouldn't `https://github.com/Bouke/django-two-factor-auth/blob/3a0fdfa191a64ded99c607e9040197803cee3f87/tests/test_views_phone.py` catch any errors in case of new incompatibilities?

moggers87:
Not sure, I wouldn't rely on it alone :smile_cat: 

There's also the changes in 4.0.0 that were discussed in https://github.com/Bouke/django-two-factor-auth/pull/326. As I understand it, that would allow invalid phone numbers that would have been previously rejected. Do our unit tests cover the case where Twilio rejects a number as invalid or were we relying on django-phonenumbers-field's "very aggressive validation". I suspect we'd be fine, but I'd rather be safe than sorry.

claudep:
Unfortunately, I'm not using phone capability in my projects at all, so I'm not in a good position for testing.

syserr0r:
> No, "fingers crossed" is not good enough. Actual user testing is required, i.e. install 5.0.0 and see if you can still set a phone number.

Sorry, just assumed when you released an update supporting Django 3.1 but your dependency doesn't that you had enough test coverage to catch issues.

If I get a chance I'll see if I can run it on our test server.

syserr0r:
Looks like we've been running 5.0.0 in staging and production for a while -- the old pip resolver warned about it but installed both anyway:

From the build log:
```
django-two-factor-auth 1.12.1 requires django-phonenumber-field<3.99,>=1.1.0, but you'll have django-phonenumber-field 5.0.0 which is incompatible.
```

From a node:
```
root@production-c88989d7d-rwb5k:/src# pip freeze | grep  -E '(two-factor|phonenumber-field)'
django-phonenumber-field==5.0.0
django-two-factor-auth==1.12.1
```

What outliers would you have me test?

kevindice:
Can confirm, 5.0.0 works in production w/ 1.13.0.

Now, since 5.0.0 removes support for the end-of-life Django 1.11 and 2.2, would it make sense for the purposes of larger project planing (@moggers87 ?) to slot this in for a major version bump of `django-two-factor-auth`?

I'm seeing that Django 1.11 support was dropped in `django-two-factor-auth` version 1.12.

A couple options:
- `django-two-factor-auth` drops support for end-of-life Django 2.2, changes the requirement to `'django-phonenumber-field>=5,<6'`, and (just going off of convention here) bumps its own version to 1.14.0 to signal a breaking change (loss of 2.2 support)
- `django-two-factor-auth` preserves official support for Django 2.2 by setting the range to `django-phonenumber-field>=4,<5` (or can allow v5 with it being the user's responsibility to pin a pre-v5 version if they really need Django 2.2 support.)

I'd propose the first option.

It appears that `django-phonenumber-field` has only dropped official support for Django 2.2, but it lives on in the test matrix: https://github.com/stefanfoulis/django-phonenumber-field/blob/master/tox.ini#L16-L17

----


tl;dr:  It'd be fine to set this to >=5 and <6 and release 1.14.0

kevindice:
More on that v4 change - `django-phonenumber-field` added aggressive validation in v2.4.0 and v4 removes this in favor of detecting invalidity in....the `str()` method.

This package isn't concerned with capturing that exception, so there's no code for us to remove there.

This issue has a bit more urgency now that pip has made package version conflicts a fatal error as of Dec 1.  That said, it's not that urgent...enjoy your holidays everyone.

If anyone needs a temp fix for your package.json: `git+git://github.com/WeVidIt/django-two-factor-auth@6a4c24b` works

moggers87:
> Now, since 5.0.0 removes support for the end-of-life Django 1.11 and 2.2, would it make sense for the purposes of larger project planing (@moggers87 ?) to slot this in for a major version bump of `django-two-factor-auth`?

This has already been resolved in #387. It was a mistake in the changelog

> This package isn't concerned with capturing that exception, so there's no code for us to remove there.

The concern is phone numbers that would have been rejected by field validation previously will now end up in the database. I've not had a chance to dive into the code to see how well we guard against that when calling Twilio or how well tested that code is.

joetheone:
Is it possible to get an update on this? 🙏

moggers87:
Updated PR title because two people have confirmed it works for them.

Also merged in latest master.

I'll merge this PR if tests pass.

atkawa7:
It seems all tests passes @moggers87. Is there a way for this to get merged

moggers87 (approved):
LGTM

Bouke:
> I would like to know why you required <3.99 and not <4 -- is this best practice or for some other reason?

When I initially wrote that, `<X` would allow pre-release versions of X. I don't know if that's still the case.

PR #409, opened 2021-02-11
Add ability to use email as totp token recipient device.

dkron:
Add ability to use email as totp token recipient device.

## Description
Implemented as Email device with similar behavior with Phone Device.

It's uses some implementation from https://github.com/Bouke/django-two-factor-auth/pull/267 . Like drift range value, etc..
But not related to user main email address because it's safer to have an option to use different email address.

On some parts I removed some `phones` context and name it `devices` in common.

Queries:
 - Is it ok to implement it this way.
 - It was developed based on last tag. Possible throttle factor make sense for EmailDevice too. Is it make sense to use different values for Phone and email in this case?
 
## Motivation and Context
It's useful to have email as option for token receive.

Other users needs it too:
https://github.com/Bouke/django-two-factor-auth/issues/404

## How Has This Been Tested?
Tested by local QA team. Released to real project about month ago.

## Types of changes

## Checklist:


dkron:
Or it should waits #331 ?

rzwink:
Can this be merged?  Looks great, and something I'd like to use.

moggers87:
I haven't had a chance to go through and do a proper review, but django-otp has its own EmailDevice now and I'd prefer to see that used than having a custom model. I'm also unsure of the necessity of the Email gateway - Django's own email sending utils already do this: https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-EMAIL_BACKEND

dkron:
Yes I didn't find that django-otp already EmailDevice, will try to implement it here.

Email gateway was added for email customization - template, optional html using, etc... It's also used Django backend by default. But because django-otp provides same options for email device it will not needed after switching to django-otp functionality.

rzwink:
Thanks for the consideration, I appreciate the support and awesome project.  Incredibly useful.

claudep:
The new PR which uses django_otp EmailDevice is now in #478, reviews and tests appreciated!

PR #413, opened 2021-03-19
Add support for Twilio Messaging Service SID

gertjanvg:
Added support to use the Twilio Messaging service as sender instead of a single caller ID.

## Description
There is an additional configuration option `TWILIO_MESSAGING_SERVICE_SID` that overrides `TWILIO_CALLER_ID` for SMS messages.

## Motivation and Context
Twilio has added support for Messaging Services (https://www.twilio.com/docs/messaging/services) which allows you to bundle multiple sender IDs into a single service. This is especially useful when you have to send messages to multiple countries, some of which have different requirements for SMS sender IDs. 

I have decided to add an extra configuration option rather than changing the way message sender IDs are used to preserve backward compatibility.

## How Has This Been Tested?
Added an extra test with the new configuration.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/gateways/twilio/gateway.py
+++ b/two_factor/gateways/twilio/gateway.py
@@ -34,6 +34,11 @@ class Twilio(object):
       Should be set to a verified phone number. Twilio_ differentiates between
       numbers verified for making phone calls and sending text messages.
 
+    ``TWILIO_MESSAGING_SERVICE_SID``
+      Can be set to a Twilio Messaging Service for SMS. This service can wrap multiple
+      phone numbers and choose one depending on the destination country.
+      When left empty the ``TWILIO_CALLER_ID`` will be used as sender ID.
+
     .. _Twilio: http://www.twilio.com/
     """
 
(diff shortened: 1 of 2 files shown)


two_factor/gateways/twilio/gateway.py
```diff
@@ -61,10 +66,16 @@ def send_sms(self, device, token):
...
+            sender_kwargs = {'messaging_service_sid': messaging_service_sid}
+        else:
+            sender_kwargs = {'from_': getattr(settings, 'TWILIO_CALLER_ID')}
+
         self.client.messages.create(
             to=device.number.as_e164,
-            from_=getattr(settings, 'TWILIO_CALLER_ID'),
-            body=body)
+            body=body,
+            **sender_kwargs)
```
moggers87, line 78: This is a bit awkward to read, why not start with `send_kwargs = {"body": body}` and then add `from_` or `messaging_service_sid` as necessary. `self.client.messages.create` can be one line then too.

moggers87 (changes requested):
One minor change requested, but other than that this looks good!

two_factor/gateways/twilio/gateway.py
gertjanvg: Fair enough, changed it up to setup the kwargs

moggers87: :+1:

moggers87 (approved):
Looks good to me!

PR #419, opened 2021-06-17
fix Phone throttling issue

### LINKED ISSUE #418: models.PhoneDevice throttling(ThrottlingMixin, TWO_FACTOR_PHONE_THROTTLE_FACTOR) is not properly implemented.
## Expected Behavior

the delay(throttling) should be imposed after successive failures in phone(SMS or call) authentication.

## Current Behavior

there's no delay(throttling).

## Possible Solution

`django_otp.models.ThrottlingMixin` says
```
    # This mixin is not publicly documented, but is used internally to avoid
    # code duplication. Subclasses must implement get_throttle_factor(), and
    # must use the verify_is_allowed(), throttle_reset() and
    # throttle_increment() methods from within their verify_token() method.
```

but in `two_factor.models.PhoneDevice`, it doesn't use `verify_is_allowed()`, `throttle_reset()`, `throttle_increment()`
```
class PhoneDevice(ThrottlingMixin, Device):
        def verify_token(self, token):  # <- doesn't have verify_is_allowed(), throttle_reset(), throttle_increment()
```
----
`django_otp.plugins.otp_email.models.EmailDevice` is a good example code
```
class EmailDevice(ThrottlingMixin, SideChannelDevice):
    def verify_token(self, token):
        verify_allowed, _ = self.verify_is_allowed()
        if verify_allowed:
            verified = super().verify_token(token)

            if verified:
                self.throttle_reset()
            else:
                self.throttle_increment()
        else:
            verified = False

        return verified
```
----
I have opened a pull request to fix this issue.
thank you.

1. set two_factor with phone authentication.
2. enter your username and password
3. enter wrong token
4. (go to 3)

kochul2000:
## Description

1. resolve #418 (models.PhoneDevice throttling(ThrottlingMixin, TWO_FACTOR_PHONE_THROTTLE_FACTOR) is not properly implemented.)
2. throttling test case has been added.
3. delays(time.sleep(x), x=2^n) are added below the test code that causes phone authentication failure and triggering throttling.

## Motivation and Context

While testing phone authentication(brute force test on authentication in several conditions), I noticed that phone authentication throttling was not working.

## How Has This Been Tested?

by `make test` and dev web site.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_views_phone.py
+++ b/tests/test_views_phone.py
@@ -1,3 +1,4 @@
+import time
 from unittest import mock
 
 from django.conf import settings
@@ -182,7 +183,9 @@ class PhoneDeviceTest(UserMixin, TestCase):
             with self.settings(TWO_FACTOR_TOTP_DIGITS=no_digits):
                 device = PhoneDevice(key=random_hex())
                 self.assertFalse(device.verify_token(-1))
+                time.sleep(1)
                 self.assertFalse(device.verify_token('foobar'))
+                time.sleep(2)  # 4, 8, 16, ..., 2^n
                 self.assertTrue(device.verify_token(totp(device.bin_key, digits=no_digits)))
 
     def test_verify_token_as_string(self):
--- a/tests/test_views_phone.py
+++ b/tests/test_views_phone.py
@@ -188,6 +188,18 @@ class PhoneDeviceTest(UserMixin, TestCase):
                 time.sleep(2)  # 4, 8, 16, ..., 2^n
                 self.assertTrue(device.verify_token(totp(device.bin_key, digits=no_digits)))
 
+    def test_throttling(self):
+        for no_digits in (6, 8):
+            with self.settings(TWO_FACTOR_TOTP_DIGITS=no_digits):
+                device = PhoneDevice(key=random_hex())
+                self.assertFalse(device.verify_token(-1))
+                time.sleep(0)  # 1
+                self.assertFalse(device.verify_token('foobar'))
+                time.sleep(0)  # 2
+                self.assertFalse(device.verify_token(totp(device.bin_key, digits=no_digits)))
+                time.sleep(4)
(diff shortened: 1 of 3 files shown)


moggers87:
Good catch! I'll do a quick review later once Travis has finished

kochul2000:
`def generate_challenge(self):`
throttling `generate_challenge` seems not working, and I'm looking for a solution to the problem.
----
thank you for comment :)

kochul2000:
```
# two_factor.models.PhoneDevice

def verify_token(self, token):
    ...
    else:
        self.throttle_increment(commit=False)
```
I found the problem with the above code. different `commit` timing between `AuthenticationTokenForm` and `DeviceValidationForm` produces below errors

If `commit=True`, throttling works normally when logging in, 
but it does not pass the `test_views_setup.py` test, and an error occurs if an incorrect code is entered when registering a phone number in setup.

If `commit=False`, throttling doesn't works normally when logging in, 
but test and setup error doesn't occur.

next code works and resolve above problems 
but I believe there's better and prettier way to fix it. 
(The code was so complex that I couldn't figure it out)
```
def verify_token(self, token):
    ...
    else:
        is_commit = True if self.id else False
        self.throttle_increment(commit=is_commit)
```

tests/test_views_phone.py
```diff
@@ -182,7 +183,21 @@ def test_verify(self):
             with self.settings(TWO_FACTOR_TOTP_DIGITS=no_digits):
                 device = PhoneDevice(key=random_hex())
                 self.assertFalse(device.verify_token(-1))
+                time.sleep(1)
                 self.assertFalse(device.verify_token('foobar'))
+                time.sleep(2)  # 4, 8, 16, ..., 2^n
```
Bouke, line 188: Introducing sleeps in the tests makes the test suite run slower. I would suggest finding an alternative. For example by using a fake clock that can be forwarded programmatically.

Bouke (changes requested):

PR #424, opened 2021-07-29
Add key to setup context

pickfire:
## Description

Add `key` to `context` within setup.

## Motivation and Context

Allows user to display the OTP with text instructions by overriding the template like
image

## How Has This Been Tested?

Added to the tests.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/views/core.py
+++ b/two_factor/views/core.py
@@ -548,8 +548,10 @@ class SetupView(IdempotentSessionWizardView):
             rawkey = unhexlify(key.encode('ascii'))
             b32key = b32encode(rawkey).decode('utf-8')
             self.request.session[self.session_key_name] = b32key
+            key = ' '.join(b32key[i:i+4] for i in range(0, len(b32key), 4))
             context.update({
-                'QR_URL': reverse(self.qrcode_url)
+                'QR_URL': reverse(self.qrcode_url),
+                'key': key,
             })
         elif self.steps.current == 'validation':
             context['device'] = self.get_device()
--- a/tests/test_views_setup.py
+++ b/tests/test_views_setup.py
@@ -35,6 +35,8 @@ class SetupTest(UserMixin, TestCase):
         self.assertContains(response, 'autocomplete="one-time-code"')
         session = self.client.session
         self.assertIn('django_two_factor-qr_secret_key', session.keys())
+        self.assertEqual(len(response.context_data['key']), 39)
+        self.assertEqual(len(response.context_data['key'].split()), 8)
 
         response = self.client.post(
             reverse('two_factor:setup'),


two_factor/views/core.py
```diff
@@ -548,8 +548,10 @@ def get_context_data(self, form, **kwargs):
             rawkey = unhexlify(key.encode('ascii'))
             b32key = b32encode(rawkey).decode('utf-8')
             self.request.session[self.session_key_name] = b32key
+            key = ' '.join(b32key[i:i+4] for i in range(0, len(b32key), 4))
```
moggers87, line 551: I don't like those spaces, what if the user needs to copy and paste the code into an app?

It might be possible to have the same effect without literal spaces via some HTML and CSS.

moggers87 (changes requested):
Good, except for the spaces. Also I think it would be worth updating the templates so this feature gets some exposure.

two_factor/views/core.py
pickfire: They can use `|cut:" "` for that (on the other hand there are no methods to easily add spaces in template). Without this it is harder to separate them out. Most sites I have seen online display it with spaces. Sticking it together makes it harder to read.

moggers87: End users can't use `|cut:" "` and going to be pretty annoyed when they can't copy and paste that code into their app because of the spaces.

moggers87: What's wrong with a template tag that groups the key chars into 4s and wraps them in `<span>` tags? That way when the user selects them to copy and paste there are no spaces *and* we could use some CSS to add padding to make the code easier to read.

pickfire: I mean developers can just put `|cut:" "` to let users see it without space. If there is a copy and paste code in the first place, shouldn't they write some javascript to do that? Also, a bunch of text without spaces is unreadable.

What's wrong with developer typing in an extra `|cut:" "`? It would be more troublesome to type in a template tag that is a few lines long if they wanted to add it. But comparing to removing it, it's just few characters. Also, it is not possible to add spacing every 4 characters using only CSS from what I know.

moggers87: HTML:
```
<div id="secret-key"><span>0123</span><span>4567</span><span>890a</span><span>bcde</span></div>
```

CSS:

```
#secret-key span {
  padding-left: 0.5em;
  padding-right: 0.5em;
}
```

A working example: https://jsfiddle.net/3x9yj8ao/1/

I don't think it would be too much effort to write a template tag that would output the required HTML for a given key.

That would meet both our requirements:

1. There are visual spaces to make the key easier to read
2. There are no literal spaces, so the key can easily be copied and pasted

pickfire: It is not much effort but compared to adding `cut:" "`, it requires significantly more effort. Or maybe I include the template tag as well? Also, even with literal spaces, it could be copy pasted. I did copy paste it as well, including the space and excluding the space does not matter.

PR #425, opened 2021-07-30
Remove Travis and housekeeping

moggers87:
A whole bunch of things that have been bugging me for a while.

## Description

- Removed Travis and replace it with Github actions
- Default branch in Django git is now main
- django-admin is preferred over django-admin.py, the latter being
  removed in a future version of Django
- Try to import url_has_allowed_host_and_scheme over is_safe_url where
  possible
- Added Python 3.9 to test matrices
- Updated CodeCov action to latest version

## Motivation and Context

Travis.org is no longer running. We could migrate to Travis.com, but we already have Github actions defined so we may as well just use them.

## How Has This Been Tested?

Github actions should run in this PR and we should see coverage reports working.

## Types of changes

## Checklist:


moggers87:
Codecov should not be blocking merging, but it is for some reason. I need to work out why and then get this merged.

WayneLambert:
@moggers87 Could it be that the `master` branch is protected and has a rule on it that requires all of its status checks to pass before it can be merged? See https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging

In the `build.yml` file, it has a setting to fail the CI if there are errors. This would mean that this would be a failing status check on the protected branch which would prevent the merge.

There is a suggestion within the Codecov docs that there is an `informational` config setting which can be used if you only want the coverage report to give information only rather than actually failing a build.

Also, looking at the GitHub action itself, [the docs](https://github.com/marketplace/actions/codecov) say that v1 will be deprecated in Feb 2022 and are recomming using v2.

Of course, I am not an administrator and therefore am unable to tell what branch rules are in place, so I might be off-target with this.

moggers87:
@WayneLambert ah, I'd somehow missed `fail_ci_if_error` being set. Thanks!

moggers87:
I guess project settings are overriding `.codecov.yml`

moggers87:
@Bouke once this has been merged, I think you'll need to go into the repo settings and remove the required Travis build as this PR moves us to Github Actions only.

WayneLambert:
> @WayneLambert ah, I'd somehow missed `fail_ci_if_error` being set. Thanks!

Hi @moggers87, You're welcome and thank you for moving this on. Much appreciated! 👍🏼

.github/workflows/build.yml
```diff
@@ -5,30 +5,13 @@ on: [push]
...
-    name: Python ${{ matrix.python-version }}, Django ${{ matrix.django-version }} (${{ matrix.variant }})
+    name: Python ${{ matrix.python-version }}
     runs-on: ubuntu-latest
-    continue-on-error: ${{ matrix.experimental == 'true' }}
     strategy:
       matrix:
-        python-version: ['3.5', '3.6', '3.7', '3.8']
-        django-version: ['2.2', '3.0', '3.1', master]
-        variant: [normal, yubikey, custom_user]
-        experimental: ['true', 'false']
```
Bouke, line 16: This changes the build matrix to only have the python versions, and using tox to build the remaining matrix within that version. I wrote it that way to have a nice overview inside Github to inspect the matrix output. On the other hand, the UI assumes smaller matrices and having such a long list can be harder to work with. What's your take on this?

See also https://github.com/Bouke/django-two-factor-auth/pull/430 that enables this workflow for PRs.

Bouke (commented):
Overall LGTM. I have some doubt about the build matrix, but am fine either way.

Coverage has 3 required checks [patch, tests, two_factor] of which only 1 succeeds at the moment. I can merge the PR ignoring these failures, but maybe you want me to disable the requirement on some of those checks?

What else do you want me to do?

moggers87:
> Coverage has 3 required checks [patch, tests, two_factor] of which only 1 succeeds at the moment. I can merge the PR ignoring these failures, but maybe you want me to disable the requirement on some of those checks?

I thought those had been changed to informational only a while ago? Maybe I'm misremembering!

In any case, I think we can ignore the coverage failures on this PR for sure.

.github/workflows/build.yml
moggers87: Personally I find the duplicated build matrix to be a bit of nightmare for project maintenance, which is why I changed it.

debdolph:
Any progress? I think this PR blocks #408 :/…

Frikster:
For those waiting on this who don't want to use this development branch, I have tested and found both the following repos [django-fido](https://github.com/CZ-NIC/django-fido) and [kagi](https://github.com/justinmayer/kagi) to work quite nicely to offer Webauth 2FA.

Bouke (approved):

PR #426, opened 2021-08-10
Changing hidden input submit to general CSS.

yopiti:
If using a different CSS framework than bootstrap this causes an empty button to appear. The fix keeps the functionality and hides it.

## Types of changes

## Checklist:


claudep:
Could you try adding the `hidden` (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) property, if check if the form Enter key behavior is kept?
Adding a direct style property might be problematic on a CSP point of view.

yopiti:
I agree. hidden is the better choice. Closing this pull request.

PR #427, opened 2021-08-11
Changing hidden input submit to facilitate hidden property not css

yopiti:
Setting field to hidden, to avoid conflict with pages that are not using Bootstrap as CSS framework

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/templates/two_factor/core/login.html
+++ b/two_factor/templates/two_factor/core/login.html
@@ -19,7 +19,7 @@
     {% include "two_factor/_wizard_forms.html" %}
 
     {# hidden submit button to enable [enter] key #}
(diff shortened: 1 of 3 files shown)


yopiti:
Can you please merge this?

moggers87:
The template has the comment `{# hidden submit button to enable [enter] key #}` above that input - does that still work with a hidden input? I can't imagine it does.

moggers87 (dismissed):
:+1:

moggers87 (changes requested):
There are other places in our templates where this same issue occurs, see https://github.com/jazzband/django-two-factor-auth/search?q=d-none&type=code

claudep (approved):

PR #434, opened 2021-10-06
Avoid backup phone query if phones can't be used

jpaniagualaconich:
`backup_phones` now returns phones only if they can be used as backup.
## Description

Phones can't be used if there's no phone gateway configured. `backup_phones` should not return devices that aren't usable as second factor backups. This PR changes the `backup_phones` function accordingly.

## Motivation and Context

This change avoids an extra query to the DB for devices that will not be used as second factors if no phone gateway is configured.

## How Has This Been Tested?

`UtilsTest.test_backup_phones` has been extended to also check that `backup_phones` returns an empty `QuerySet` when no phone gateway is configured.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -143,15 +143,28 @@ class UtilsTest(UserMixin, TestCase):
 
 class PhoneUtilsTests(UserMixin, TestCase):
     def test_backup_phones(self):
-        self.assertQuerysetEqual(list(backup_phones(None)),
-                                 list(PhoneDevice.objects.none()))
+        gateway = 'two_factor.gateways.fake.Fake'
         user = self.create_user()
         user.phonedevice_set.create(name='default', number='+12024561111')
         backup = user.phonedevice_set.create(name='backup', number='+12024561111')
-        phones = backup_phones(user)
 
-        self.assertEqual(len(phones), 1)
-        self.assertEqual(phones[0].pk, backup.pk)
+        parameters = [
+            # with_gateway, with_user, expected_output
+            (True, True, [backup.pk]),
+            (True, False, []),
+            (False, True, []),
+            (False, False, [])
+        ]
+
+        for with_gateway, with_user, expected_output in parameters:
+            gateway_param = gateway if with_gateway else None
+            user_param = user if with_user else None
+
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -29,15 +29,23 @@ def test_default_device(self):
...
+            self.assertQuerysetEqual(list(backup_phones(None)),
+                                    list(PhoneDevice.objects.none()))
+            phones_with_gw = backup_phones(user)
+
+            self.assertEqual(len(phones_with_gw), 1)
+            self.assertEqual(phones_with_gw[0].pk, backup.pk)
+
+        phones_without_gw = backup_phones(user)
+
+        self.assertEqual(len(phones_without_gw), 0)
```
moggers87, line 48: This took me way too long to work out what was actually being tested here. These should probably be three separate tests for each of the conditions we're testing here, but I'd accept a short comment above each `backup_phones`

two_factor/utils.py
```diff
@@ -15,6 +15,9 @@ def default_device(user):
 
 
 def backup_phones(user):
+    if (getattr(settings, 'TWO_FACTOR_CALL_GATEWAY', None) is None
+            and getattr(settings, 'TWO_FACTOR_SMS_GATEWAY', None) is None):
+        return PhoneDevice.objects.none()
     if not user or user.is_anonymous:
         return PhoneDevice.objects.none()
```
moggers87, line 22: Could this be one if statement or do you think that will make it difficult to read and parse what is happening?

moggers87 (changes requested):
Just a few little changes and you also need to resolve some conflicts :+1:

tests/test_utils.py
jpaniagualaconich: i rewrote it as a parameterised test (it's in the "new" place `tests.test_utils.PhoneUtilsTests.test_backup_phones`)

two_factor/utils.py
jpaniagualaconich: i rewrote this too. Let me know what you think.

moggers87 (approved):
Looks good :+1:

Issue #438, opened 2021-11-24
Remove upper limit to qrcode dependency


claudep:
Currently, django-two-auth-factor has an upper limit to the qrcode version. In a project where I use qrcode-7.3.1, this causes a downgrade of the dependency in the project (or a conflict if my project setup a hard dep on qrcode > 7).
Could that upper limit be dropped, please?

Bouke:
The reason for having an upper limit is to prevent breakage if a backwards-incompatible version were to be released. A major version bump often indicates such backwards-incompatible changes when adhering semver. So I'd suggest upping the limit if there's no breakage.

PR #443, opened 2021-12-28
Update django-phonenumber-field to newer version

### LINKED ISSUE #439: Support django-phonenumberfield 6
Hi,

please update the dependency for `django-phonenumberfield` and upload a new release to PyPI.

Thanks!

debdolph:
Closes #439 

## Description
Allow django-phonenumber-field version <7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Changed
 - Suppressed default_app_config warning on Django 3.2+
-- qrcode dependency limit upped to 7.99.
+- qrcode dependency limit upped to 7.99 and django-phonenumber-field to 7
 
 ## 1.13.1
 
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         'Django>=2.2',
         'django_otp>=0.8.0',
         'qrcode>=4.0.0,<7.99',
-        'django-phonenumber-field>=1.1.0,<6',
+        'django-phonenumber-field>=1.1.0,<7',
         'django-formtools',
     ],
     extras_require={


Bouke:
Thanks. What, if any, backward-incompatible changes were made that might impact this project?

claudep:
That question can be asked for any dependency. I think the strategy should be different whether the project is a lib used by other projects or an "end-user" project. For the latter, it seems like best practices recommend to pin down dependencies as precisely as possible (pip-lock and such) so as to not have surprises when deploying in production.

For libraries, I think the situation is different, because if you are too strict in dependency pinning, you enter what is call the dependency hell (typically when two dependencies of your projects pin sub-dependencies with conflicting specifications). That happened to me in #438. That's why my opinion is to only specify minimal versions in that case (see for example install_requires in Django setup.cfg, https://github.com/django/django/blob/main/setup.cfg). I think most Django reusable apps choose this strategy. Then, it should be the role of the test suite to catch incompatibilities with new dependency versions. Some projects even choose to run their test suite with -Werr, so as any deprecation warnings are immediately caught by the test suite and you have less risks of encountering breakages with new dependency versions.

Natureshadow:
Now that all the Jazzband stuff has happened and there is so much activity on the issue tracker, can we finally expect a PyPI upload?

Bouke:
> For libraries, I think the situation is different, because if you are too strict in dependency pinning, you enter what is call the dependency hell (typically when two dependencies of your projects pin sub-dependencies with conflicting specifications).

See the recent breakage with Django 4.0. We had some other breakages in the past as well, which I why I'm not a big fan of this. I don't think specifying an upper limit for a "known good major version" qualifies as "too strict", but I can see your point in reaching a state where you cannot update other dependencies as those require unsupported versions from this version.

> That's why my opinion is to only specify minimal versions in that case

Well now you have another problem; some dependency releases a new major version with changed behaviour. This project hasn't checked against this new version and your app breaks in some subtle way. Maybe it results in a situation where 2fa can be bypassed, who knows. Is that a risk we should accept? Each dependency increases the amount of baby-sitting required, especially when there's no upper bound. @claudep would you like to accept this project having to baby-sit all dependencies?

Natureshadow:
> Well now you have another problem; some dependency releases a new major version with changed behaviour. This project hasn't checked against this new version and your app breaks in some subtle way. Maybe it results in a situation where 2fa can be bypassed, who knows. Is that a risk we should accept? Each dependency increases the amount of baby-sitting required, especially when there's no upper bound. @claudep would you like to accept this project having to baby-sit all dependencies?

That is what semantic versioning was invented for. Limiting the
dependency to ^5.2 (which is >=5.2,<6.0) helps.

I agree that new major versions should not be pulled in
automatically. If the major version is bumped, breakage is guaranteed,
because this is why it is bumped.

Bouke:
> Now that all the Jazzband stuff has happened and there is so much activity on the issue tracker, can we finally expect a PyPI upload?

@Natureshadow Your comment comes over as entitled and disrespectful of the free time contributed by the contributors.

claudep:
> > For libraries, I think the situation is different, because if you are too strict in dependency pinning, you enter what is call the dependency hell (typically when two dependencies of your projects pin sub-dependencies with conflicting specifications).
> 
> See the recent breakage with Django 4.0. We had some other breakages in the past as well, which I why I'm not a big fan of this. I don't think specifying an upper limit for a "known good major version" qualifies as "too strict", but I can see your point in reaching a state where you cannot update other dependencies as those require unsupported versions from this version.
> 
> > That's why my opinion is to only specify minimal versions in that case
> 
> Well now you have another problem; some dependency releases a new major version with changed behaviour. This project hasn't checked against this new version and your app breaks in some subtle way. Maybe it results in a situation where 2fa can be bypassed, who knows. Is that a risk we should accept? Each dependency increases the amount of baby-sitting required, especially when there's no upper bound. @claudep would you like to accept this project having to baby-sit all dependencies?

It depends on your definition of "baby-sitting" :-)

To be clear, I think both strategies have merits and drawbacks. For me, whatever the strategy about pinning, the key is proper maintainership and good test coverage. If those criteria are met, breakages will be detected soon and fixed quickly, and any upper dependency limit can also be checked with the test suite and upped quickly (I guess that's the problem currently, that is an upper limit not upped for too long).

Typically for this PR, if the test suite passes, we should trust it and commit such proposals ASAP.

Bouke:
> To be clear, I think both strategies have merits and drawbacks. For me, whatever the strategy about pinning, the key is proper maintainership and good test coverage. If those criteria are met, breakages will be detected soon and fixed quickly, and any upper dependency limit can also be checked with the test suite and upped quickly (I guess that's the problem currently, that is an upper limit not upped for too long).

Proper maintainership and good test coverage is key to any strategy, of course. The passing of time doesn't magically make the dependency's new version compatible. I agree that not being able to update dependencies in your app / downstream project can be annoying. Having the project under the jazzband umbrella can help with that, allowing the community to step in and contribute.

> Typically for this PR, if the test suite passes, we should trust it and commit such proposals ASAP.

Feel free to approve the PR and merge it 👍.

Natureshadow:
> Your comment comes over as entitled and disrespectful of the free time contributed by the contributors.

Aha.

Sorry, but I am a free software contributor as well, and a significant amount of my volunteer time has gone down the drain due to this very library being badly maintained and outdated on PyPI all the time.

Problems happen, and normally I make pull requests to fix issues and donate money to upstream maintainers. But that one thing, releasing to PyPI, can only be done by a few people, and if these people miss out on uploading patches, all the other work is quite useless.

claudep (approved):
Release note completed, looks good.

PR #448, opened 2021-12-30
Add pip cache dir to speed up tests on GitHub

claudep:


Bouke:
Comparing [this run](https://github.com/jazzband/django-two-factor-auth/runs/4665474212?check_suite_focus=true) with [some other](https://github.com/jazzband/django-two-factor-auth/runs/4665775032?check_suite_focus=true) I cannot see a speed up. Total time is 7m 22s vs 6m 35s. Am I missing something, or is there something wrong?

claudep:
No, this run cannot be faster as this is the first to use the cache (and the keys are different for each Python version, so you won't get speed-ups "intra-run", but only between several runs).

Bouke:
So with your update I can see the cache step being run and also downloading some cached packages. The speed-up is negligible however; the install step goes from ~12s to ~6s but downloading the cache takes almost the speed gain (~4s). I would expect GitHub to run a pypi mirror, so it doesn't hit the (fast) CDNs anyway. So is adding this complexity to the build worth it to maybe speed up the builds by a few percentage?

claudep:
Indeed, the gain for dtfa is minor. I guess we may keep it for a bit less pressure on PyPI itself. I'm fine if you'd prefer dropping this.

claudep:
Looks like there is now a simpler way to activate the cache: https://github.com/jazzband/tablib/commit/94ffe67e5

PR #458, opened 2022-01-03
When validating TOTP in TOTPDeviceForm, allow a drift in clocks of +/-1 instead of just -1

### LINKED ISSUE #455: Bug in loop over tolerance in TOTPDeviceForm?
In the clean_token function of TOTPDeviceForm in two_factor/forms.py. there is the following line (163):

    for offset in range(-self.tolerance, self.tolerance):

self.tolerance is hardcoded to 1, which means this loop will range over the values -1, 0.  Did you intend it to range over -1, 0, 1?  If this is the case, the line should be

    for offset in range(-self.tolerance, self.tolerance+1):

I find that, without this, I cannot complete the 2FA activation when using Google Authenticator to scan the QR code.

There is a similar line in two_factor/models.py (90):

        for drift in range(-5, 1):

This will range over the values (-5, -4, -3, -2, -1, 0).  Personally, this does not create an issue for me but the one in clean_token does.

If you don't want to change this, is it possible to make tolerance a settable parameter?  I can see now way of changing it in my code without a monkey patch.

mbakereth:
## Description

When validating a TOTP, the TOTPDeviceForm class allows for a clock drift defined by the instance variable tolerance, hardcoded to 1.  While looping of values within this tolerance, the loop started at -tolerance but ended at tolerance-1 instead of tolerance.

A test class is also added.  It sets the time in the TOTP class to a constant value (to make tests deterministic), and confirms the TOTP is accepted when there is an offset of -1, 0 or 1 between the clocks and that it is not accepted when the offset is -2 or +2.

## Motivation and Context
I found when registering and authenticator for the first time, I could only very rarely get the initial TOTP to validate until I made this change.

The change makes the loop the same as in [this example](https://python.hotexamples.com/examples/django_otp.oath/TOTP/drift/python-totp-drift-method-examples.html).

Closes #455 

## How Has This Been Tested?
New test class added as described above.  This confirms it works for the offset values it should work for, and not for one value either side.
All existing tests continue to succeed.  Tested with Python 3.9 on Mac.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_totpdeviceform.py
+++ b/tests/test_totpdeviceform.py
@@ -1,5 +1,3 @@
-import unittest
-from unittest.mock import patch, MagicMock
 from binascii import unhexlify
 import time
 
@@ -32,7 +30,7 @@ class TOTPDeviceFormTest(UserMixin, TestCase):
         self.old_TOTP_init = old_TOTP_init
         def new_TOTP_init(self, key, step=30, t0=0, digits=6, drift=0):
         	old_TOTP_init(self, key, step, t0, digits, drift)
-        	self._time = TEST_TIME
+        	self.time = TEST_TIME
         django_otp.oath.TOTP.__init__ = new_TOTP_init
 
     def restore_time_for_totp(self):
--- a/tests/test_totpdeviceform.py
+++ b/tests/test_totpdeviceform.py
@@ -1,12 +1,7 @@
-import time
 from binascii import unhexlify
 
 import django_otp.oath
-from django import forms
-from django.conf import settings
-from django.shortcuts import resolve_url
 from django.test import TestCase
-from django.urls import reverse
 
 from two_factor.forms import TOTPDeviceForm
 
@@ -16,6 +11,7 @@ from .utils import UserMixin
 # as 3 Jan 2022
 TEST_TIME = 1641194517
 
+
 # This class test how the TOTPDeviceForm validator handles drift between its clock
 # and the TOTP device.  If there is a drift in the range [-tolerance, +tolerance],
 # where tolerance is hardcoded to 1, then the TOTP should be accepted.  Outside this
@@ -24,13 +20,14 @@ class TOTPDeviceFormTest(UserMixin, TestCase):
     key = '12345678901234567890'
 
     def fix_time_for_totp(self):
-    	# Override constructor for TOTP class to set a fixed time.  This will
-    	# prevent it from calling time(), which would make the unittest unpredictable
+        # Override constructor for TOTP class to set a fixed time.  This will
+        # prevent it from calling time(), which would make the unittest unpredictable
         old_TOTP_init = django_otp.oath.TOTP.__init__
         self.old_TOTP_init = old_TOTP_init
+
         def new_TOTP_init(self, key, step=30, t0=0, digits=6, drift=0):
-        	old_TOTP_init(self, key, step, t0, digits, drift)
-        	self.time = TEST_TIME
+            old_TOTP_init(self, key, step, t0, digits, drift)
+            self.time = TEST_TIME
         django_otp.oath.TOTP.__init__ = new_TOTP_init
 
     def restore_time_for_totp(self):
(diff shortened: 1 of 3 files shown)


claudep:
Thanks! Cuuld you please fix the isort/flake8 errors (see pre-commit.ci details).

tests/test_totpdeviceform.py
```diff
@@ -0,0 +1,78 @@
...
+# This class test how the TOTPDeviceForm validator handles drift between its clock
+# and the TOTP device.  If there is a drift in the range [-tolerance, +tolerance],
+# where tolerance is hardcoded to 1, then the TOTP should be accepted.  Outside this
+# range, it should be rejected.
+class TOTPDeviceFormTest(UserMixin, TestCase):
+    key = '12345678901234567890'
+
+    def fix_time_for_totp(self):
+        # Override constructor for TOTP class to set a fixed time.  This will
+        # prevent it from calling time(), which would make the unittest unpredictable
```
claudep, line 24: Instead of monkey-patching, I think it would be a bit more elegant to use mock for this. You should be able to mock `django_otp.oath.time` to return your test value. To be explored...

tests/test_totpdeviceform.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Use this as the Unix time for all TOTPs.  It is chosen arbitrarily
+# as 3 Jan 2022
+TEST_TIME = 1641194517
+
+
+# This class test how the TOTPDeviceForm validator handles drift between its clock
+# and the TOTP device.  If there is a drift in the range [-tolerance, +tolerance],
+# where tolerance is hardcoded to 1, then the TOTP should be accepted.  Outside this
+# range, it should be rejected.
```
claudep, line 18: I would rather define this part as the docstring of the `TOTPDeviceFormTest`  class.

tests/test_totpdeviceform.py
```diff
@@ -0,0 +1,78 @@
...
+class TOTPDeviceFormTest(UserMixin, TestCase):
```
claudep, line 19: I'm not sure that `UserMixin` is required here, did you test removing it?

tests/test_totpdeviceform.py
```diff
@@ -0,0 +1,78 @@
...
+        device_totp = self.totp_with_offset(1)
+        self.form = TOTPDeviceForm(TOTPDeviceFormTest.key, None,
+            data={'token': device_totp})
+        self.assertTrue(self.form.is_valid())
+
+    def test_offset_minus2(self):
+        device_totp = self.totp_with_offset(-2)
+        self.form = TOTPDeviceForm(TOTPDeviceFormTest.key, None,
+            data={'token': device_totp})
+        self.assertFalse(self.form.is_valid())
```
claudep, line 72: A little bonus would be to test the form error message (same below).

claudep (changes requested):
Globally good, thanks! Some minor testing things that could be improved in my opinion.

claudep:
Could you please also add a line in the CHANGELOG?

mbakereth:
All added:
- Docstring instead of comment
- @patch (class level) instead of constructor override
- UserMixin dependency removed
- Test error message
- CHANGELOG updated

tests/test_totpdeviceform.py
```diff
@@ -0,0 +1,70 @@
...
+    and the TOTP device.
+
+    If there is a drift in the range [-tolerance, +tolerance], (tolerance is hardcoded
+    to 1), then the TOTP should be accepted.  Outside this range, it should be rejected.
+    """
+
+    key = '12345678901234567890'
+
+    def restore_time_for_totp(self):
+        django_otp.oath.TOTP.__init__ = self.old_TOTP_init
```
claudep, line 27: This method is no longer used.

claudep (commented):
We're almost there! You may squash your commits after the last required change.

claudep (approved):
Thank you again!

PR #459, opened 2022-01-08
updated classifiers

anukaal:
added support for python 3.10


claudep:
Thanks a lot for your contribution, but this was done as part of #464.

PR #461, opened 2022-01-16
Prepare release of django-two-factor-auth 1.13.2

claudep:


moggers87 (approved):
Looks fine.

I assume translations are up to date (I don't know if you can push new strings to Transifex yet, but anyone can pull in new translations)

I don't know if you have permissions to do a PyPI release, if not I can do it.

claudep:
I don't think strings change much between 1.13.1 and 1.13.2, so I wouldn't worry too much about that.
As for the release, we should start to rely on Jazzband release method. As soon as we push a new tag, the release process should start. See https://jazzband.co/about/releases

claudep:
Just pushed 1.13.2 tag. I think all project leads should soon receive an email with instructions.

PR #463, opened 2022-01-18
Removed ignored flake8 errors

claudep:


setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 exclude = build,dist,docs,.git,.tox,./two_factor/migrations/*.py,./two_factor/views/__init__.py,./tests/migrations/*
-ignore = E123,E128,E265,E266,E402,E501,E731,W503,W504,W601
+ignore = W503,W504
```
moggers87, line 3: We should probably fix those warnings too, but it can wait.

moggers87 (approved):
:+1:

PR #464, opened 2022-01-18
Dropped Python 3.5/3.6 and added Python 3.10

claudep:
Python 3.5 security support ended 13 Sep 2020
Python 3.6 security support ended 23 Dec 2021


moggers87 (approved):
:+1:

PR #466, opened 2022-01-27
Cache default device onto user object

### LINKED ISSUE #465: Duplicate query for LoginView
## Expected Behavior

`LoginView` should not have extra duplicated and unnecessary queries.

## Current Behavior

Each step in `LoginView` seemed to be calling `has_token_step` and `has_backup_view`, and both calls `default_device` for the same thing but the results are not cached which results in duplicated queries.

## Possible Solution

Cache `has_token_step` and `has_backup_view` or maybe have cached version of `default_device`. At least we get this in django debug toolbar.

image

1. Use `LoginView`, I can't really test examples in two-factor-auth out so not sure how to test out a smaller version, so not sure if it's just us or if it's intended.

## Context

Reduce the login loading time, but still the improvements probably not that significant, at least from my side I can see ~32ms improvements probably if there is a fix.

## Your Environment

* Browser and version: firefox 96
* Python version: 3.8
* Django version: 3.1
* django-otp version: 1.1.3
* django-two-factor-auth version: 1.13.1
* Link to your project: closed source

moggers87:
Reduces database queries during login

Fixes #465

## Description

- For logged in users, cache the default 2FA device onto the user object
- Anonymous users are unaffected as they don't cause any database queries

## Motivation and Context

`default_device` is called multiple times during various views in django-two-factor-auth. In particular, it appears to be the cause of the excessive database queries found in #465 

## How Has This Been Tested?

Added to tests already present

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/utils.py
+++ b/two_factor/utils.py
@@ -5,12 +5,17 @@ from django_otp import devices_for_user
 
 from two_factor.models import PhoneDevice
 
+USER_DEFAULT_DEVICE_ATTR_NAME = "_default_device"
+
 
 def default_device(user):
     if not user or user.is_anonymous:
         return
+    if hasattr(user, USER_DEFAULT_DEVICE_ATTR_NAME):
+        return getattr(user, USER_DEFAULT_DEVICE_ATTR_NAME)
     for device in devices_for_user(user):
         if device.name == 'default':
+            setattr(user, USER_DEFAULT_DEVICE_ATTR_NAME, device)
             return device
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ### Added
 - Python 3.10 support
 
+### Changed
+- default_device utility function now caches the found device on the given user object
+
 ### Removed
 - Python 3.5 and 3.6 support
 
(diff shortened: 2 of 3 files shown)


two_factor/utils.py
```diff
@@ -5,12 +5,17 @@
 
 from two_factor.models import PhoneDevice
 
+USER_DEFAULT_DEVICE_ATTR_NAME = "_default_device"
```
claudep, line 8: What is the interest here to put the attribute name in a constant? Who would customize it?

moggers87: I feel like it's a bit of a magic number if we don't put it in a constant.

> Who would customize it?

We would when we discover there's some other project that attaches data to the user object and they just happen to use `_default_device` too.

claudep: OK, is the PR in Draft status for a reason? Do you plan a Changelog entry?

moggers87: Yeah, I just put this out real quick so there was a patch available. I will add a changelog entry and then mark it as ready for review :+1:

claudep (approved):

PR #467, opened 2022-01-28
Removed not_skip isort config as it is deprecated in isort 5

claudep:
From the 5.0.0 Changelog:
Since there is no longer composition negative form settings (such as --dont-skip
or it's config file variant not_skip) are no longer required and have been removed.


moggers87 (approved):
:+1:

PR #468, opened 2022-01-29
Make two_factor.plugins a package

claudep:
Follow up to 07dac39d357836.

Looks like an `__init__.py` file was missing from the commit adding the new plugin architecture.


moggers87:
I'd have thought tests would have failed. I guess not!

moggers87 (approved):

PR #470, opened 2022-01-29
Add a registry-based system for plugins

claudep:
This is a follow-up to the 07dac39d3, where we try to delegate much of the code specific to particular plugins to the plugins code. There might be some more efforts to do to achieve this final goal, but I think this is already a good step forward.


claudep:
@moggers87 I would love to get your input on this. I'll gladly fix the code coverage part if the general idea is approved.

moggers87 (commented):
This looks really promising :+1:

moggers87:
I have requested a review from @moggers87

claudep:
Oh, that's a good friend of yours? :-)

claudep:
@moggers87 I can totally understand that you can't find the time to review this, or more generally to make this package evolve.
Just tell me, and I'll try to find other reviewers.

two_factor/plugins/phonenumber/__init__.py
```diff
@@ -0,0 +1,4 @@
+import django
+
+if django.VERSION <= (3, 2):
```
moggers87, line 3: Is this line (and the one for yubikey) necessary? I.e. will Django 4.0 start spitting out warnings if we don't put in this if statement

moggers87 (approved):
One question, but but approved either way.

This is a great improvement and certainly makes the prospect of adding new methods less daunting (from a maintenance point of view at least)

moggers87:
> @moggers87 I can totally understand that you can't find the time to review this, or more generally to make this package evolve. Just tell me, and I'll try to find other reviewers.

Sorry about the wait, I wanted to find the time/energy to give this a proper review. This is some good work :smile_cat:

moggers87 (changes requested):
Ha! Spoke too soon: `two_factor.plugins.yubikey.utils` has zero test coverage and I think you just forgot to remove that file.

two_factor/plugins/phonenumber/__init__.py
claudep: Yes, it is. See 1b6b6a3f9 for the equivalent commit for the main `two_factor` app.
See https://github.com/django/django/blob/stable/4.0.x/django/apps/config.py#L171 for the location where Django raises the warning.

claudep:
Ah yes, good catch! I removed that file now.

moggers87 (approved):

PR #471, opened 2022-01-30
Removed double decoration of LoginView

claudep:
LoginView.dispatch is already decorated with method_decorator's.


moggers87 (approved):
:+1:

moggers87:
This PR is an excellent example of why code coverage should not block merging :smile_cat:

PR #475, opened 2022-03-04
Email 2fa

claudep:
New iteration of the patch initially from @Atterratio, then @pickfire, now rebased on latest master. Reviews welcome!


claudep:
I see now in a parallel PR discussion that `django_otp` also have an `EmailDevice`. So I guess this PR should be reworked to consider using that device.

claudep:
The new PR which uses django_otp EmailDevice is now in #478, reviews and tests appreciated!

PR #476, opened 2022-03-04
Also cache get_device for SetupView

claudep:
get_device() result was already cached for LoginView, but not for SetupView.


moggers87 (approved):
LGTM :+1:

PR #479, opened 2022-03-05
OTP tokens are digits, not integers

claudep:
Co-authored-by: "Laurent Mignon (ACSONE)" <laurent.mignon@acsone.eu>

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -3,9 +3,10 @@ from django.contrib.auth import get_user_model
 from django.shortcuts import resolve_url
 from django.test.utils import TestContextDecorator
 from django_otp import DEVICE_ID_SESSION_KEY
+from django_otp.oath import totp
 
 from two_factor.plugins.registry import registry
-from two_factor.utils import default_device
+from two_factor.utils import default_device, totp_digits
 
 
 class UserMixin:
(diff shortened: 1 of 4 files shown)


tests/utils.py
```diff
@@ -56,3 +57,7 @@ def enable(self):
 
     def disable(self):
         registry._methods = self.old_methods
+
+
+def totp_str(key):
+    return str(totp(key)).zfill(6)
```
moggers87, line 63: Minor nitpick: we should call `totp_digits()`, otherwise we'll get unnecessary test failures if `TWO_FACTOR_TOTP_DIGITS` ever changes.

moggers87 (changes requested):
Just a minor nitpick and one question.

- Will this break if for some reason a user has written their own form HTML and has hard coded the token field to be a numeric input? If that is the case, we might want to highlight that in the changelog when we next do a release.

tests/utils.py
claudep: Good catch, fixed.

claudep:
No, nothing in this patch would make a form with `IntegerField` fail. It's just that using an `IntegerField` is risky as possibly the form would clean the token value and strip any initial `0`. In my opinion, a token is never an integer, it's a string composed of digits.

moggers87 (approved):
Approved

moggers87:
As far as I can tell, Codecov is failing because it believes `totp_str` is not covered by tests, despite there being tests that call `totp_str` :upside_down_face: I'm going to merge anyway.

PR #480, opened 2022-03-09
fix submitting Back button on pressing Enter

PetrDlouhy:
## Description
If I press Enter on the two factor form, the Back button gets pressed. This fixes that.

Snímek obrazovky_2022-03-09_17-18-33

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/templates/two_factor/_wizard_actions.html
+++ b/two_factor/templates/two_factor/_wizard_actions.html
@@ -1,5 +1,6 @@
 {% load i18n %}
 
+<button type="submit" style="display: none"></button>
 {% if cancel_url %}
   <a href="{{ cancel_url }}"
      class="float-right btn btn-link">{% trans "Cancel" %}</a>


two_factor/templates/two_factor/_wizard_actions.html
```diff
@@ -5,7 +5,7 @@
      class="float-right btn btn-link">{% trans "Cancel" %}</a>
 {% endif %}
 {% if wizard.steps.prev %}
-  <button name="wizard_goto_step" type="submit"
+  <button name="wizard_goto_step" type="button"
```
claudep, line 8: If you do that, clicking the button won't going back, unless you add a JS handler that do something when the button is clicked.

claudep (changes requested):

two_factor/templates/two_factor/_wizard_actions.html
dopry: @PetrDlouhy do you have time or capacity to address @claudep's change request?

PetrDlouhy: OK, @claudep is actually right. And it seems, that the first button is always default in HTML: https://stackoverflow.com/questions/1963245/multiple-submit-buttons-on-html-form-designate-one-button-as-default

So if we don't want to colide with user defined form styling, I am duplicating the submit button as hidden to the first place in the form.

PetrDlouhy:
This seems to be working already with the current master. I am closing this PR.

Issue #186, opened 2017-01-04
make voice token format customizable


patroqueeet:
acc to https://github.com/Bouke/django-two-factor-auth/blob/f308883d7e1e327f06080da254c75a20e5fcc4fb/two_factor/gateways/twilio/views.py#L83 there is always a dot in between. in german language this changes the way the numbers are spoken by the voice: `vier`(4) becomes `vierte` which does confuse the user listening. it would be great if the `.` can be configured or other ways of making the string understood would be used.

Bouke:
Indeed, that code is written like that so it will change a number "1234" to "1. 2. 3. 4.". The goal was to have the number "spelled out" instead of read like a number. E.g. "1, 2, 3, 4" instead of "one thousand two hundred thirty four". This worked for the languages I tested with (nl, en). I understand that for other languages it might not work. Can you check what a good solution would be for German and possibly for English as well?

patroqueeet:
hey, I checked your code and even twilio api docu says that `.,` is interpreted as a natural pause. but when I tested today I listed to the behavior documented above. I would recommend to make the char a setting so that users can either play around and even try `-` or similar things e.g. ` and ` to make this perfect for their language. alternatively twilio recommends the `<pause>` tag, but I have no experience with that and do not know how and if that works.

Bouke:
You can see the XML that Twilio is reading [in the tests](https://github.com/Bouke/django-two-factor-auth/blob/9ea5da23f2b90a32630171b3c48bc33256d9199d/tests/test_gateways.py#L39).

```
<?xml version="1.0" encoding="UTF-8" ?>'
<Response>
  <Say language="en">Your token is 1. 2. 3. 4. 5. 6. Repeat: 1. 2. 3. 4. 5. 6. Good bye.</Say>
</Response>
```

From the documentation on `Pause`, I expect the following should work:

```
<?xml version="1.0" encoding="UTF-8" ?>'
<Response>
  <Say language="en">Your token is 1<Pause/>2<Pause/>3<Pause/>4<Pause/>5<Pause/>6. Repeat: 1<Pause/>2<Pause/>3<Pause/>4<Pause/>5<Pause/>6. Good bye.</Say>
</Response>
```

Given that a Pause is at least 1 second, this might be a very slow prompt. That's why I would like to get some experience with a solution that would work for all languages, instead of language-specific workarounds.

patroqueeet:
hey, reasonable and understood. next steps? how can I support?

[...]

PR #484, opened 2022-03-18
make Twilio token separator configurable #186

patroqueeet:
for different countries the `.` separator will create invalid language used by Twilio. e.g. German `4.` -> `vierte` (fourth) but should be `vier` (four).

## Types of changes

## Checklist:

tests didn't exec at all here... sry.


moggers87 (changes requested):
Tests?

claudep:
I wonder if the proper solution could be to output `<Pause>` tags between `<Say>` tags for each individual digit.

```
<Response>
   <Say language="de">1</Say>
   <Pause>
   <Say language="de">4</Say>
   <Pause>
   ...
</Response>
```

What do you think? This way, it wouldn't depend on the language used.

patroqueeet:
Hm, don't know. would need to test it with German language. Will do next week...

did your read https://github.com/jazzband/django-two-factor-auth/issues/186#issuecomment-270358640 ?

patroqueeet:
@claudep I did check your proposal and found it to be working for German (at least). I cannot judge if this would be a proper solution for English language. furthermore it would have to be fed into `'token': _('Your token is %(token)s. Repeat: %(token)s. Good bye.'),` which is as of a now a markup-less string, suitable for translation. if we split this up, we would have to adjust all translations... I managed to get all working until I found out I had to mess with translations. as of now I'd be happy got get one of the core devs help here, so this will match all your specs.

claudep:
Here's an alternative patch using the `<Pause>` tag: #488. Indeed, it affects translatable strings, but we shouldn't refrain improving things because of that.

patroqueeet:
@claudep want me to check for German language?

dopry:
@patroqueeet could you submit the docker compose setup as a separate PR? I would be interested in that.

patroqueeet:
> @patroqueeet could you submit the docker compose setup as a separate PR? I would be interested in that.

https://github.com/jazzband/django-two-factor-auth/pull/498/files :)

PR #497, opened 2022-05-07
feat: drop django 2.2, 3.0, and 3.1 support

dopry:
Drop support for EOL Django versions 2.2, 3.0, and 3.1

## Motivation and Context
As an open-source team, we don't have the resources to maintain compatibility with end-of-life code.  Consumers who need to work with older versions of Django can use older releases. 

## Checklist:
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
+# Changelog
+
 ## Unreleased
 
 ### Added
 - Enforcing a redirect to setup of otp device when none available for user [#550](https://github.com/jazzband/django-two-factor-auth/pull/500)
+
+### Removed
+- Django 2.2, 3.0, and 3.1 support
 
 ## 1.14.0
 
--- a/README.rst
+++ b/README.rst
@@ -37,8 +37,8 @@ providing Django sessions with a foreign key to the user. Although the package
 is optional, it improves account security control over
 ``django.contrib.sessions``.
 
-Compatible with modern Django versions. At the moment of writing that's
-including 2.2, 3.1, 3.2, and 4.0 on Python 3.7, 3.8, 3.9 and 3.10.
+Compatible with supported Django and Python versions. At the moment of writing that
+includes 3.2 and 4.0 on Python 3.7, 3.8, 3.9 and 3.10.
 Documentation is available at `readthedocs.org`_.
 
 
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -3,8 +3,7 @@ Requirements
 
 Django
 ------
-Modern Django versions are supported. Currently this list includes Django 2.2,
-3.0, 3.1, 3.2, and 4.0.
+Supported Django versions are supported. Currently this list includes Django 3.2, and 4.0.
 
 Python
 ------
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ setup(
     license='MIT',
     packages=find_packages(exclude=('example', 'tests')),
     install_requires=[
-        'Django>=2.2',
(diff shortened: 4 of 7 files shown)


.coveragerc
```diff
@@ -17,3 +17,5 @@ exclude_lines =
     # Don't complain if tests don't hit defensive assertion code:
     raise AssertionError
     raise NotImplementedError
+# skip empty files, they seem to get misreported on codecov
+skip-empty = true
```
moggers87, line 21: This config line appears to have caused CodeCov to report missing coverage on `tests/__init__.py`. We either need to configure CodeCov to also skip empty files or remove this line and put up with empty files in our coverage reports.

moggers87 (changes requested):
Looks good, just one changed needed to fix coverage issues.

claudep:
As a release might come soon, I would suggest to wait a bit and merge this just after the release.

.coveragerc
dopry: Here is coverage without skip-empty
```
@@            Coverage Diff             @@
##           master     #497      +/-   ##
==========================================
- Coverage   98.42%   98.34%   -0.08%     
==========================================
  Files          55       58       +3     
  Lines        2475     2606     +131     
  Branches      266      274       +8     
==========================================
+ Hits         2436     2563     +127     
- Misses         24       26       +2     
- Partials       15       17       +2     
```

The diff is 100% covered, I feel like the -0.08% is an error in  the change in coverage caldulation in this case.

claudep:
I don't trust codecov results for a long time now... don't bother.

.coveragerc
```diff
@@ -16,4 +16,4 @@ exclude_lines =
 
     # Don't complain if tests don't hit defensive assertion code:
     raise AssertionError
-    raise NotImplementedError
+    raise NotImplementedError
```
claudep, line 19: Avoid unrelated changes in this patch, please.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 ### Removed
 - Python 3.5 and 3.6 support
+- Django 2.2, 3.0, and 3.1 support
```
claudep, line 16: Please move this above (in a new Unreleased section), please.

two_factor/plugins/yubikey/__init__.py
```diff
@@ -1,4 +0,0 @@
-import django
-
-if django.VERSION <= (3, 2):
-    default_app_config = 'two_factor.plugins.yubikey.apps.TwoFactorYubikeyConfig'
```
claudep, line 4: I think there are more django.VERSION checks in master.

claudep (changes requested):
Now that 1.14.0 is out, we can go forward with this.

CHANGELOG.md
dopry: fixed.  dealing with CHANGELOG conflicts isn't very contributor friendly. It would be nice if these were a part of the release process, either manually aggregated by the maintainers or by something like [semantic-release](https://python-semantic-release.readthedocs.io/en/latest/).

two_factor/plugins/yubikey/__init__.py
claudep: It may be that the comparison is wrong and should have been `< (3, 2)`. Could you check?

claudep (approved):
Thanks!

claudep:
@moggers87 We need your review to go forward as you requested changes.

moggers87 (approved):

PR #498, opened 2022-05-09
Docker dev env

patroqueeet:
## Description
simple docker compose setup to run tests

## Motivation and Context
easier dev start

## How Has This Been Tested?
by using it :)

## Types of changes
config files

## Checklist:
- [ x My change requires a change to the documentation.


dopry:
@patroqueeet so this simplifies dev startup by not requiring the developer have python installed locally?

patroqueeet:
> @patroqueeet so this simplifies dev startup by not requiring the developer have python installed locally?

I have a Mac and at least to me, it helped prevent any kind of local setup. check the code out, run the tests via docker and be happy :)

up to you if you consider it helpful. was at least to me

patroqueeet:
@dopry this has been added now. **NOTE**: the example app is broken with Djv4. You'd need to add `"Django>=2.2,<4",` to your setup.py :)

CONTRIBUTING.rst
```diff
@@ -32,6 +39,10 @@ covering all supported Python and Django version with::
 
     tox
 
+With docker run tests::
+
+    docker-compose run --rm cli bash -c "pip install -r requirements_dev.txt; make test"
```
moggers87, line 44: Is this really easier than running `pip install --user tox; tox`? If Docker supports one-shot services maybe we should have one to run the tests?

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3.9"
```
moggers87, line 1: Is this the Python version to be used? If not, where is that defined?

moggers87 (commented):
Just a few comments. I'm not sure what this adds to the project and I'm skeptical that this will be well maintained in the future (I don't think any of the current devs use Docker regularly, but maybe I'm wrong?)

CONTRIBUTING.rst
patroqueeet: that is always a matter of personal flavour. I like to prevent messing around locally with dozens of different virtualenvs. but I'm not the maintainer. I offered what I ran locally and others asked for it. so its up to you to find the right way for this lib.

docker-compose.yml
patroqueeet: rtfm, sry.

dopry:
> @dopry this has been added now. **NOTE**: the example app is broken with Djv4. You'd need to add `"Django>=2.2,<4",` to your setup.py :)

I'll take a look later today. :)

Issue #499, opened 2022-05-12
Hitting otp_required as a user without device results in infinite login loop


MWeesenaar:
During implementation of this module, I noticed that in some use cases, I got stuck in an infinite login loop.

## Expected Behavior
For step 4, I would expect a message stating that the page the user tries to access requires OTP and without the user having an OTP device, that this is not possible. OR) the user is being redirected to the OTP setup page, and after setting up the OTP, have the possibility to get back to the page where the user initially wanted to go to.
Note that this might be troublesome for body-related requests (e.g. POST/PUT), but I have the feeling (not confirmed from my side) that this is already the case for the current implementation.

## Current Behavior
When going to an `otp_required`-view, after signing in with a user without an OTP device, the user gets stuck in an infinite login-loop.

## Possible Solution
I would say that an else-clause could be useful in [core.py#L160](https://github.com/jazzband/django-two-factor-auth/blob/master/two_factor/views/core.py#L160).
I will create a pull request with the suggested change and updated tests.
The only question I have, right now, is the following: since this change will introduce a 'breaking change' (in the sense of user-flow), please advise whether it should be configureable or not. I tend to say no, since this feels like a user-workflow related bug; but I prefer the maintainers to have an opinion on this.

## Steps to Reproduce (for bugs)
0. Install the module as per instructions, configure appropriately and make sure you got atleast one user without an otp-device linked and `@otp_required` view - for making sure that it is not related to my Django instance, I could reproduce it on a clean-freshly installed Django project
1. Make sure you're not logged in on the Django app and go to the otp_required-view
2. Validate that you're being redirected to `/account/login/?next={view_url}`
3. Log in with your user credentials
4. You are again, directed to the `/account/login/` page, with the same `?next={view_url}`

## Context
It affects me because I have a few dozen users of my app, who are now in a loop of logging in, coming to me, and I need to explain them to go to the OTP setup page.

## Your Environment
* Browser and version: Chrome 103.0.5052.0 (Official Build) canary (arm64) 
* Python version: 3.10 (and initially found on 3.6)
* Django version: 4.0.4 (and reproducible on 3.2.12)
* django-otp version: 1.1.3
* django-two-factor-auth version: 1.13.2
* Link to your project: N/a

alstr:
I'm getting this problem even when I have a device set up.

claudep:
This should hopefully be fixed in master with 4bd592cf6543

PetrDlouhy:
[...]

PR #491, opened 2022-05-05
feat: redirect admins to two_factor:setup if  two_factors are required a two factor is not enabled for the account 

dopry:
I'm picking up the work from #370 in this PR since it has seemingly gone stale. 

## Description
Currently, if OTP is set to required for the admin interface and a user does not have admin privileges. Logging in fails without any feedback.  This PR modifies the login process to redirect admins to setup OTP instead. 

## Motivation and Context
This change is required because there is a dead end in the Login UX.  It applies to #219.

## How Has This Been Tested?
I am resubmitting the PR as a WIP at this juncture. 

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #500, opened 2022-05-13
Enforcing a redirect to setup of otp device when none available for user

MWeesenaar:
This pull request will enforce a user, without OTP device, to be redirected to the OTP setup page keeping the redirected-from in mind.

## Description
Please refer to issue #499 for more details and context.

## Motivation and Context
Please refer to issue #499 for more details and context.

## How Has This Been Tested?
I created a scenario test: go to a protected page as anonymous user, log in, setup OTP and validate whether you get the option to be redirected to the page you meant to visit.
The current change has impact on the existing flow for users without OTP devices configured.

Currently I have not made any. I might be able to create a short movie about the workings.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have added tests to cover my changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+### Added
+- Enforcing a redirect to setup of otp device when none available for user [#550](https://github.com/jazzband/django-two-factor-auth/pull/500)
+
 ## 1.14.0
 
 ### Added
--- a/tests/test_views_login.py
+++ b/tests/test_views_login.py
@@ -38,7 +38,7 @@ class LoginTest(UserMixin, TestCase):
         response = self._post({'auth-username': 'bouke@example.com',
                                'auth-password': 'secret',
                                'login_view-current_step': 'auth'})
-        self.assertRedirects(response, resolve_url(settings.LOGIN_REDIRECT_URL))
+        self.assertRedirects(response, reverse('two_factor:setup'))
 
         # No signal should be fired for non-verified user logins.
         self.assertFalse(mock_signal.called)
@@ -80,7 +80,8 @@ class LoginTest(UserMixin, TestCase):
             {'auth-username': 'bouke@example.com',
              'auth-password': 'secret',
              'login_view-current_step': 'auth'})
-        self.assertRedirects(response, redirect_url, fetch_redirect_response=False)
+        self.assertEqual(self.client.session.get('next'), redirect_url)
+        self.assertRedirects(response, reverse('two_factor:setup'), fetch_redirect_response=False)
 
     def test_valid_login_with_disallowed_external_redirect(self):
         redirect_url = 'https://test.disallowed-success-url.com'
@@ -90,7 +91,7 @@ class LoginTest(UserMixin, TestCase):
             {'auth-username': 'bouke@example.com',
              'auth-password': 'secret',
              'login_view-current_step': 'auth'})
-        self.assertRedirects(response, reverse('two_factor:profile'), fetch_redirect_response=False)
+        self.assertRedirects(response, reverse('two_factor:setup'), fetch_redirect_response=False)
 
     @mock.patch('two_factor.views.core.time')
     def test_valid_login_primary_key_stored(self, mock_time):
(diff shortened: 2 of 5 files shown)


mikkiweesenaar:
Optionally the documentation about [Enforcing two-factor](https://github.com/jazzband/django-two-factor-auth/blame/master/docs/implementing.rst#L47) _could_ be updated or extended. I leave that up to the maintainers to determine whether that would be helpful, and if so, I will be pleased to do so.

As I cannot put this pull request out of draft[1], could one of the maintainers mark this pull request as ready for review?


[1] `Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can mark a draft pull request as ready for review.`

MWeesenaar:
Optionally the documentation about [Enforcing two-factor](https://github.com/jazzband/django-two-factor-auth/blame/master/docs/implementing.rst#L47) _could_ be updated or extended. I leave that up to the maintainers to determine whether that would be helpful, and if so, I will be pleased to do so.

As I cannot put this pull request out of draft[1], could one of the maintainers mark this pull request as ready for review?


[1] `Only those with write access to this repository can mark a draft pull request as ready for review.`

dopry:
@MWeesenaar Does this conflict with #491? I think in #491 I was automatically redirecting the user if next was set. This code looks like it waits until the "setup complete" view and then requires the user to click a link to get back to their intended destination.  I kind of prefer saving the user the action if we know a destination.  Do you think there is a good reason to require the user to take another action is this point?

mikkiweesenaar:
Hi @dopry ,

Thanks for your message. Please consider the scenario I described in #499. It seems, that the redirection works too well when you do not have MFA setup yet.
I do not think that #491 really conflicts with this PR; where you solved it for Admin specifically where my main focus was to fix the login-loop.
I can validate my scenario with your branch/PR to validate if your solution fixes the problem I was facing; would that help?

dopry:
@MWeesenaar I was addressing the same issue, but for admins. They ended up in the same scenario you are describing in #499. I think they're compatible. We're both redirecting to the setup, but from different entry points in the UX Flow. We both intercept and redirect to OTP Setup. The only conflict I saw was handling the ?next  after the user completes the setup.  I feel like we should reconcile the UX flows of our implementations. I'd rather not create churn by implementing one flow and then immediately changing it with the final behavior dependent on the order of merges. 

This is how your implementation in behaves

```mermaid
flowchart TD
   START((user requests\notp required page))-->CHECK_OTP{Is the user\nOTP Verified}
   CHECK_OTP-->|YES|NEXT[Requested Page is Delivered]
   CHECK_OTP-->|NO|OTP_SETUP[OTP Setup Page]
   OTP_SETUP-->|Complete OTP SETUP|OTP_SETUP_COMPLETE[OTP Setup  Complete]
   OTP_SETUP_COMPLETE-->|User clicks line to goto requested page|NEXT
```

This is how my implementation behaves, but only on the admin.

```mermaid
flowchart TD
   START((user requests\notp required page))-->CHECK_OTP{Is the user\nOTP Verified}
   CHECK_OTP-->|YES|NEXT[Requested Page is Delivered]
   CHECK_OTP-->|NO|OTP_SETUP[OTP Setup Page]
   OTP_SETUP-->|Complete OTP SETUP|NEXT
```

My implementation removes the additional user interaction of needing to click the next link on the SetupComplete view in favor of redirecting the user once the setup is complete.   This is modeled on the same flow as a login intercept, where the user is automatically returned to their intended destination after login.

What is your reason for sending the user to the Setup Complete page and asking them to perform an additional interaction?

MWeesenaar:
Thanks for your feedback @dopry !
I must say that I do not have a very clear answer to your question why I think that manual interaction is better than redirecting automatically.
Would you suggest me removing the manual interaction?

dopry:
Let's update it to automate the redirect. I worry that users will get further distracted from what they are doing if we add additional interactions to get them back to their original task.

MWeesenaar:
I understand your point of view.
I will try to fix this somewhere this weekend. As you indicated, the code is there to do so, so it should not take too much time to actually implement it. :)

dopry:
@MWeesenaar I rebased your PR on master. I updated to use RedirectURLMixin in place of SuccessURLAllowedHostsMixin for Django 4.1 compatibility that had been added upstream.

dopry (approved):
I've tested locally and this looks good to me. It even works for the admin. It looks like it will supersede my work in #491.

PR #488, opened 2022-04-23
Use <Pause> tag when saying token digits

claudep:

PR #502, opened 2022-05-16
German translations #488

patroqueeet:
## Description
German translations for #488


claudep:
Thanks @patroqueeet, but you should update the strings in Transifex (https://www.transifex.com/Bouke/django-two-factor-auth), then we will download all updated strings before the next release.

Issue #503, opened 2022-05-16
Add white background to QR image for browsers in dark mode


jcu721:
My computer preference theme is set to Dark. Django interprets this nicely, but the QR code shows up as dark on dark and is very difficult to see. Can be read by some QR readers but not by others. 

## Expected Behavior
The QR image should always have a white background instead of transparent. 
Browsers with dark mode enabled that are overriding the base template should be able to have a readable QR image that can be scanned easily. 

## Current Behavior
When dark mode is enabled, the background is set to black and the transparent image shows up directly on the default black background. It can be read by one phone I tested with, but not the 1Password browser QR scanner. It is also very difficult to even visually tell there is a QR code there.  
<img width="750" alt="Screen Shot 2022-05-16 at 6 26 25 PM" src="https://user-images.githubusercontent.com/2374425/168692457-a92cd9cc-4531-47ca-a335-069fbd58c21d.png">

## Possible Solution
Hardcode the style background to be white on the img tag. 
`<p><img src="{{ QR_URL }}" alt="QR Code" style="background:white"/></p>`

This results in an easy to see QR image with a nice margin/border:
<img width="175" alt="Screen Shot 2022-05-16 at 6 28 16 PM" src="https://user-images.githubusercontent.com/2374425/168693315-51dcfc67-7078-46b0-a732-1173b5bbc66a.png">

## Steps to Reproduce (for bugs)
1. Have a custom _base.html template that extends the base django template: `{% extends "admin/base.html" %}`
2. Have computer (mac, in my case) theme set to dark (System Preferences -> General -> Appearance)
3. Navigate through 2FA setup to get to QR page
4. Try to scan with in browser QR scanner

## Context
The dark on dark QR code can actually be read by a phone (tested with iphone 13 mini), but cannot be read by the 1Password in-browser QR scanner. I'd imagine it would also work poorly with older phones / certain cameras. 

## Your Environment
* Browser and version: Firefox 100.0
* Python version: 3.7.7
* Django version: 3.2.13
* django-otp version: 1.1.3
* django-two-factor-auth version: 1.13.2

claudep:
@jcu721, do you plan to work on a patch?

jcu721:
@claudep  just opened a PR, https://github.com/jazzband/django-two-factor-auth/pull/504, let me know if I need to do any further testing.

PR #504, opened 2022-05-23
Update QR img style to always have a white background

jcu721:
## Description
Changed QR image background to always be white. 

## Motivation and Context
Solves an issue where the QR code image is displayed on a browser with dark mode enabled and is unable to be read. 
This happens when the base template is overridden and another base template that respects dark/light mode is used (for ex. the [default django admin template](https://docs.djangoproject.com/en/4.1/ref/contrib/admin/#theming-support)). 

Fixes https://github.com/jazzband/django-two-factor-auth/issues/503

Existing bad behavior this PR is addressing: https://user-images.githubusercontent.com/2374425/169890814-8cc38b67-9549-4c47-81f1-e2606a8ea5a0.mov

## How Has This Been Tested?
Tested manually with firefox, chrome and safari, against light mode and dark mode on MacOS Big Sur. 

Tested by using in a django admin installation with an overridden template and with the default template. 

Existing tests all still pass.  

## Screenshots:
With custom template (Bootstrap + original Django admin styling):
Dark mode:
Screen Shot 2022-05-23 at 2 07 05 PM
Light mode:
Screen Shot 2022-05-23 at 2 07 31 PM

The default template does not distinguish between dark mode and light mode, so remains a full white background no matter what the system preferences are set to:
https://user-images.githubusercontent.com/2374425/169885203-835eb269-9f03-4622-9a95-9e98e1017c4d.mov

## Types of changes

## Checklist:

Not sure how I would add a unit test for this? Didn't find any snapshot level tests for UI changes in the existing tests.


claudep (approved):
Thanks for the patch!

PR #505, opened 2022-05-28
Updated LoginView inheritance after Django change

claudep:
Related to https://github.com/django/django/commit/12576bd371.


moggers87 (approved):
:+1:

Issue #501, opened 2022-05-15
Release 1.14.0


claudep:
I'd like to make a 1.14.0 release soon. @moggers87, ok for that? I'll prepare a commit with Changelog and such in the coming days.

claudep:
I just pushed the updated translatable strings to Transifex.

claudep:
#507 is the PR for the Changelog, but #505 and #506 should be merged first, then the translations should be pulled from Transifex.

moggers87:
Pulled the latest translations from Transifex. PR here: https://github.com/jazzband/django-two-factor-auth/pull/508

Once that's been merged we can tag and upload a release

claudep:
Crap, I tagged and pushed 1.14.0, then realized that we missed bumping version in some places. So I reverted the tag and prepared #509.

moggers87:
Tagged a new release and I think we just wait for CI to do the upload now, right?

moggers87:
Done and done: https://pypi.org/project/django-two-factor-auth/1.14.0/

Thanks @claudep for taking the lead on getting a release out :smiley_cat:

claudep:
:champagne:

PR #508, opened 2022-06-06
Latest translations pulled from Transifex

moggers87:
Required for #501


claudep (approved):
:+1: (even if I hate the new plural equation for French from Transifex, but nothing we can change)

Issue #473, opened 2022-02-15
Valid backup authentication codes don't log the user in.


miniatureweasle:
Backup authentication codes don't log the user in.

## Expected Behavior

Submitting a valid backup authentication code in the login form should result in a successful sign-in, with the user being redirected out of the login page.

## Current Behavior

After submitting the valid code, the API responds with 200 but the user remains on the sign-in page, instead of being logged in.

## Steps to Reproduce (for bugs)

1. Generate a backup authentication code from the provided setup wizard
2. Store these backup codes for later
2. Logout, and open the login page
3. When prompted to use an authentication code, click the button to use the backup codes instead
4. Enter a valid backup code, instead of a successful login, the page is reloaded.

## Context

We are wanting to use this package to provide 2FA to our customers

## Your Environment

* Browser and version:
* Python version: 3.9.5
* Django version: 3.1.8
* django-otp version:
* django-two-factor-auth version: 1.13.2 (latest)
* Link to your project: http://hiddenapp.com/

moggers87:
> the API responds with 200

On success, you should get a 302. Are you using your own templates? It's possible there's an error but your template isn't displaying it.

Gautier:
I have received the same bug report as @miniatureweasle for one of my Django websites and tracked the issue to the backup device `throttling_failure_count` being incremented every time the user logs in.

I have raised a failing test under https://github.com/jazzband/django-two-factor-auth/pull/521 to demonstrate the issue as a unit test.

The problem seems to come from the fact that `AuthenticationTokenForm` tries to verify the token against each device and if the backup device is tried before the phone device a correct token received by SMS will increment the `throttling_failure_count` of the backup device. The user is able to log in correctly but after enough logins the backup device `throttling_failure_count` will be high enough for the backup tokens to be effectively unusable.

https://github.com/jazzband/django-two-factor-auth/blob/830915a8f002cdf44dccb20faca60ce4a48d90c9/two_factor/forms.py#L126-L132

## Workaround

My workaround is a custom `AuthenticationTokenForm._chosen_device`. This makes the form try only the token against the device relevant to the current step (initially the default phone device, but the user can choose an alternative phone device or a backup token). 

```python
 class SiteAuthenticationTokenForm(AuthenticationTokenForm):
    def _chosen_device(self, user):
        return self.initial_device
```

[...]

PR #521, opened 2022-07-26
Only validate token against chosen device (#473)

Gautier:
## Description

Changes the AuthenticationTokenForm to only validate a token against the chosen device.

The first version of this PR only includes a failing test to demonstrate the problem and the second version is a suggestion on how to fix it.

## Motivation and Context

I received the same error report as #473 for one of my system using django-two-factor-auth and noticed the backup tokens `throttling_failure_count` were being incremented even though the phone TOTP worked.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [X] My code follows the code style of this project. (no black?)
- [X] I have added tests to cover my changes.

### DIFF
--- a/two_factor/forms.py
+++ b/two_factor/forms.py
@@ -3,7 +3,9 @@ from time import time
 
 from django import forms
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.translation import gettext_lazy as _
+from django_otp import devices_for_user
 from django_otp.forms import OTPAuthenticationFormMixin
 from django_otp.oath import totp
 from django_otp.plugins.otp_totp.models import TOTPDevice
@@ -113,6 +115,7 @@ class AuthenticationTokenForm(OTPAuthenticationFormMixin, forms.Form):
         'pattern': '[0-9]*',  # hint to show numeric keyboard for on-screen keyboards
         'autocomplete': 'one-time-code',
     })
+    device_id = forms.CharField(widget=forms.HiddenInput(), required=False)
 
     # Our authentication form has an additional submit button to go to the
     # backup token form. When the `required` attribute is set on an input
@@ -123,16 +126,21 @@ class AuthenticationTokenForm(OTPAuthenticationFormMixin, forms.Form):
     use_required_attribute = False
     idempotent = False
 
+    error_messages = {
+        'invalid_device_id': _('Device id is not valid.'),
+    }
+
     def __init__(self, user, initial_device, **kwargs):
         """
-        `initial_device` is either the user's default device, or the backup
-        device when the user chooses to enter a backup token. The token will
-        be verified against all devices, it is not limited to the given
-        device.
+        `initial_device` is either the user's default device a backup device
+        when the user chooses to enter a backup token.
         """
         super().__init__(**kwargs)
         self.user = user
         self.initial_device = initial_device
+        if initial_device:
+            self.fields['device_id'].initial = initial_device.persistent_id
+        self.device_cache = None
 
         # Add a field to remeber this browser.
         if getattr(settings, 'TWO_FACTOR_REMEMBER_COOKIE_AGE', None):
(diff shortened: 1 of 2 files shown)


two_factor/forms.py
```diff
@@ -152,6 +160,22 @@ def __init__(self, user, initial_device, **kwargs):
...
+        if self.data.get("device_id"):
+            try:
+                for user_device in devices_for_user(self.user):
+                    if user_device.persistent_id == self.data["device_id"]:
+                        self.device_cache = user_device
+                        break
+            except ObjectDoesNotExist:
+                raise forms.ValidationError(self.error_messages['invalid_device_id'])
+
+    def _chosen_device(self, user):
```
moggers87, line 173: This method doesn't appear to be called by anything, what's it for?

two_factor/forms.py
```diff
@@ -152,6 +160,22 @@ def __init__(self, user, initial_device, **kwargs):
                 label=label
             )
 
+    def clean_device_id(self):
+        if self.data.get("device_id"):
+            try:
+                for user_device in devices_for_user(self.user):
```
moggers87, line 166: Given that we know the device ID, couldn't we just fetch it directly rather than iterating over all devices like this?

two_factor/forms.py
```diff
@@ -123,16 +126,21 @@ class AuthenticationTokenForm(OTPAuthenticationFormMixin, forms.Form):
...
+    }
+
     def __init__(self, user, initial_device, **kwargs):
         """
-        `initial_device` is either the user's default device, or the backup
-        device when the user chooses to enter a backup token. The token will
-        be verified against all devices, it is not limited to the given
-        device.
+        `initial_device` is either the user's default device a backup device
+        when the user chooses to enter a backup token.
```
moggers87, line 136: Not sure I understand this new comment

moggers87 (commented):
Just some comments/questions after a quick read through

two_factor/forms.py
Gautier: I have updated the comment to include the crucial missing word "or" in "default device or a backup device"

two_factor/forms.py
Gautier: `_chosen_device` is the method called by the `OTPTokenForm._clean_otp` method in django-otp.

It currently returns None which makes django-otp try to match the token against each user's devices in order until one works (in `match_token`) but since `match_token` uses `devices_for_token` that iterates over the devices that way `for model in device_classes():` there is no guarantee that it won't try the backup devices first.

I have added more details in my comment at the related issue at : https://github.com/jazzband/django-two-factor-auth/issues/473#issuecomment-1196384536 under the heading "Full investigation".

two_factor/forms.py
Gautier: Good point, I have updated the PR to use `Device.from_persistent_id` instead.

milafrerichs:
Any updates on this? We ran into this issue as well. Would be great to have a fix on the main package. 
Can I help in any way?

PetrDlouhy:
@moggers87 @Gautier What is state of this PR? Could I help somehow?

claudep:
Rebasing looks like the next step, then @PetrDlouhy, your review would be welcome, too!

PetrDlouhy:
I have tried to rebase this in #683, but I am not able to resolve the review comments there without much deeper investigation.

@Gautier Would you be able to look at it?
I am a bit confused, because I tried to run the `tests.test_views_login.LoginTest.test_totp_token_does_not_impact_backup_token` test with the unfixed `forms.py` code and it seems to pass.

I will try to give it more time, but it would save me some nerves if you could look at it.

PetrDlouhy:
I did investigate it a bit more and I realized that in the new version of `django-two-factor-auth` there is `AuthenticationTokenForm._chosen_device()` newly implemented.
I am not yet sure if that is enough to fix #473 or there is some case in which it could still fail.

PetrDlouhy:
The relevant test from this PR is merged now and is passing because of other fixes in the current branch.

Other changes included in this PR are probably not needed and if they are, they would need a clear test case first.
I am closing this, open new PR in such cases.

Issue #519, opened 2022-07-25
`phone_device` table isn't created when using `--nomigrations`


klistwan:
I have a 2.2 Django project with `django-otp==0.8.0`, `django-two-factor-auth[phonenumbers]==1.14.0`, and an app `two_factor_auth` that makes use of these two libraries.

When tests are run with migrations, they pass:

```zsh
$ ./manage.py test two_factor_auth
  Applying two_factor.0001_initial... OK
  Applying two_factor.0002_auto_20150110_0810... OK
  Applying two_factor.0003_auto_20150817_1733... OK
  Applying two_factor.0004_auto_20160205_1827... OK
  Applying two_factor.0005_auto_20160224_0450... OK
  Applying two_factor.0006_phonedevice_key_default... OK
  Applying two_factor.0007_auto_20201201_1019... OK
----------------------------------------------------------------------
Ran 13 tests in 5.549s

OK
```

The migrations that are run are the ones from here: https://github.com/jazzband/django-two-factor-auth/tree/master/two_factor/migrations

But when we run tests without migrations, the tests fail:

```zsh
$ ./manage.py test two_factor_auth --nomigrations
django.db.utils.OperationalError: no such table: two_factor_phonedevice
```

I went back through our commit history and can confirm this issue started appearing when we:
- upgraded `django-otp` from 0.7.5 to 0.8.0
-  upgraded `django-two-factor-auth` from 1.9.1. to 1.14.0
- added `two_factor.plugins.phonenumber` as an INSTALLED_APP
- added `django-two-factor-auth[phonenumbers]==1.14.0` as a dependency

## Expected Behavior
We would expect running `$ ./manage.py test two_factor_auth --nomigrations` to also have all tests pass.

## Current Behavior
When tests are run without migrations (as is done in CI and local environments because our project has hundreds of migrations), they fail.

```
$ ./manage.py test two_factor_auth --nomigrations

Traceback (most recent call last):
  File "/opt/miniconda3/lib/python3.8/site-packages/two_factor/views/profile.py", line 38, in get_context_data
    'default_device': default_device(self.request.user),
  File "/opt/miniconda3/lib/python3.8/site-packages/two_factor/utils.py", line 16, in default_device
    for device in devices_for_user(user):
  File "/opt/miniconda3/lib/python3.8/site-packages/django_otp/__init__.py", line 80, in devices_for_user
    for device in model.objects.devices_for_user(user, confirmed=confirmed):
[...]

PR #527, opened 2022-08-20
Move Phone model to the correct app while keeping data

mick88:
## Description
Fixes issue #519 by adding state migration to the phone plugin app. This MR moves the model to the same app where model code resides.
Note that all operations in this MR are purely state migrations and no database operations are carried out therefore there is no risk of data loss. The table name is hardcoded as to avoid renaming database table following the model being moved to a different app.

## Motivation and Context
As documented in issue #519, django does not correctly create the phone model as the model code is in a different app from where the model is assigned to. 

## How Has This Been Tested?
Migrations were ran. Running makemigrations shows that there are no outstanding model changes.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


two_factor/migrations/0008_delete_phonedevice.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.4 on 2022-08-20 19:03
```
claudep, line 1: I think you could safely remove those Generated lines, in both files.

mick88: done

claudep:
Could you still reformat the new initial migration so flake8 is happy (line length <= 119)?

claudep (approved):
Thanks!

PR #529, opened 2022-08-22
Add support for Django 4.1

pickfire:
## Types of changes

## Checklist:

There are some warnings on
```
RemovedInDjango50Warning: The "default.html" templates for forms and formsets will be removed. These were proxies to the equivalent "table.html" templates, but the new "div.html" templates will be the default from Django 5.0. Transitional renderers are provided to allow you to opt-in to the new output style now. See https://docs.djangoproject.com/en/4.1/releases/4.1/ for more details
```
But I can't find any `default.html` in the project so I guess it is fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Enforcing a redirect to setup of otp device when none available for user [#550](https://github.com/jazzband/django-two-factor-auth/pull/500)
+- Django 4.1 support
 
 ### Removed
 - Django 2.2, 3.0, and 3.1 support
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 - Enforcing a redirect to setup of otp device when none available for user [#550](https://github.com/jazzband/django-two-factor-auth/pull/500)
+- Django 4.1 support
```
claudep, line 7: I think I'd write "Confirmed Django 4.1 support" to signal that no code was needed to add that support. What do you think?

pickfire: I don't have any opinion on this but I can make the change if you want to.

claudep (approved):
:+1:

PR #534, opened 2022-09-03
gitignore .vscode/

jpaniagualaconich:
Added `.vscode/` to `.gitignore` so configuration files created/modified by VSCode don't show up among the modified ones.

## Motivation and Context

This change makes it easier for people to use VSCode to contribute to the project.

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ example/settings_private.py
 .eggs/
 
 .idea/
+.vscode/


claudep:
My personal approach of this is that .gitignore should only contain stuff related to everyone using the code, as stated in https://git-scm.com/docs/gitignore:

> Which file to place a pattern in depends on how the pattern is meant to be used.

> *   Patterns which should be version-controlled and distributed to other repositories via clone (i.e., files that all developers will want to ignore) should go into a .gitignore file.

> *   Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the $GIT_DIR/info/exclude file.

> *   Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

So if we follow this policy, we should rather remove `.idea` from the file. Opinions?

WhyNotHugo:
As I mentioned on another [related comment](https://github.com/jazzband/django-two-factor-auth/pull/408#discussion_r962155325):

> Hint: Put this kind of ignores in `$HOME/.config/git/ignore`. They'll apply to any repository on which you operate locally, and you won't need to add IDE-specific ignores to every single repo you work on.

Adding this kind of ignore rule to each repo is unsustainable; every single repository out there would need to have ignore for all IDE, OSs, etc (also applies for stuff like `.Thumbs.db` or `.DS_Store`, etc).

FWIW, vscode should really stop littering their config files inside code repositories; most IDEs and code editors have moved away from doing this since it means that your checkout is permanently dirty.

claudep:
++1 to what @WhyNotHugo said.

jpaniagualaconich:
@WhyNotHugo we all agree.

claudep:
Anyone of you want to make the PR removing .idea?

PR #540, opened 2022-09-15
Avoid hard reference to generator

jpaniagualaconich:
There's a hard reference to the 'generator' method as a choice. Use the first one available instead.
## Motivation and Context

It makes it possible to have a different method as the default.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/forms.py
+++ b/two_factor/forms.py
@@ -14,14 +14,16 @@ from .utils import totp_digits
 
 class MethodForm(forms.Form):
     method = forms.ChoiceField(label=_("Method"),
-                               initial='generator',
                                widget=forms.RadioSelect)
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.fields['method'].choices = [
+
+        method = self.fields['method']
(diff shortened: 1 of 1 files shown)


two_factor/forms.py
```diff
@@ -14,14 +14,16 @@
...
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.fields['method'].choices = [
+
+        method = self.fields['method']
+        method.choices = [
             (m.code, m.verbose_name) for m in registry.get_methods()
         ]
+        method.initial = method.choices[0][0]
```
claudep, line 26: I'm not sure, but did you check if setting `initial` is really needed here or if the first radio will be selected by default?

jpaniagualaconich: ok, i went and double checked. It's needed, otherwise there's no default.

This isn't a bad thing per se but I didn't want to change current behaviour.

claudep: OK, fine, thanks for checking.

claudep (approved):

PR #543, opened 2022-09-17
Make migration files pep8-compatible

claudep:


joshuadavidthomas (approved):
Nit: TBH, I don't like running linters/code formatters on migrations.

That aside, this looks fine to me.

claudep:
> Nit: TBH, I don't like running linters/code formatters on migrations.

I can understand. However that's the direction Django is taking since Django 4.1 (https://github.com/django/django/commit/d113b5a837f7). Thanks a lot, I see now that Jazzband members approval is sufficient to allow pushing patches, nice!

PR #546, opened 2022-09-26
fix pressing "Use Backup Tokens" if invalid token is already entered

### LINKED ISSUE #545: Login page usability - HTML5 validation prevents clicking on "Use Backup Token"
There are few problems with default login page usability.

## Expected Behavior

Only the "Next" button triggers form validation.

## Current Behavior
Both "Use Backup Token" or "Back" buttons can't be clicked when user enters invalid code, because of HTML5 field validation:
Snímek obrazovky_2022-09-26_12-20-21

## Possible Solution

I have solved this in my project by creating second form containing only the "Use Backup Token" and placing it on bottom of the page (which I consider better, because it doesn't disrupt the Enter 2FA form):
```htmldjango
    {% if backup_tokens %}
       <hr/>
       <small>
       <form method="post" action="" class="bkit-registration__form">
           {% csrf_token %}
            <p>{% trans "As a last resort, you can use a backup token:" %}</p>
            <p>
                <button name="wizard_goto_step" type="submit" value="backup"
                    class="btn btn-sm btn-secondary btn-block">{% trans "Use Backup Token" %}</button>
            </p>
       </form>
       </small>
    {% endif %}

```

But this doesn't solve the problem with the "Back" button.

1. Configure 2FA with security codes.
2. Log in
3. Enter code "xyzabc"
4. Pres "Use Backup Token" or "Back"

## Context
This might be problem especially if user did enter the backup code into the 2FA code field - if he expects that the code could be entered in this field, he will get very confused possibly unable to change his mind model to be able to delete the code and click on the button again.

## Your Environment

* Browser and version: Chrome Verze 102.0.5005.115
* Python version: 3.10.4
* Django version: 4.0.7
* django-otp version: 1.1.3
* Link to your project: www.blenderkit.com

PetrDlouhy:
Fix #545,
I added blank block `form_extra_params` to both forms to allow easy setting of i.e. additional classes in overridden form.

## Motivation and Context
Described in #545 

## How Has This Been Tested?
I have tested this manually since this could be tested only by Selenium or similar tests.

Snímek obrazovky_2022-09-26_18-41-17

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/templates/two_factor/core/login.html
+++ b/two_factor/templates/two_factor/core/login.html
@@ -31,14 +31,23 @@
         </button>
       {% endfor %}</p>
     {% endif %}
-    {% if backup_tokens %}
-      <p>{% trans "As a last resort, you can use a backup token:" %}</p>
-      <p>
-        <button name="wizard_goto_step" type="submit" value="backup"
-                class="btn btn-secondary btn-block">{% trans "Use Backup Token" %}</button>
-      </p>
-    {% endif %}
 
     {% include "two_factor/_wizard_actions.html" %}
   </form>
+
+  {% block 'backup_tokens' %}
+    {% if backup_tokens %}
+       <hr>
+       <div class="backup_tokens_form">
+       <form action="" method="post">
+           {% csrf_token %}
+            <p>{% trans "As a last resort, you can use a backup token:" %}</p>
+            <p>
+                <button name="wizard_goto_step" type="submit" value="backup"
+                    class="btn btn-sm btn-secondary btn-block">{% trans "Use Backup Token" %}</button>
+            </p>
+       </form>
+       </div>
+    {% endif %}
+  {% endblock %}
 {% endblock %}


claudep:
Thanks for the PR. The `form_extra_params` part is not directly related to this issue, so please remove it from this PR and explain your need and use case in another ticket.

PetrDlouhy:
@claudep OK, fixed.

two_factor/templates/two_factor/core/login.html
```diff
@@ -31,14 +31,23 @@ <h1>{% block title %}{% trans "Login" %}{% endblock %}</h1>
...
-                class="btn btn-secondary btn-block">{% trans "Use Backup Token" %}</button>
-      </p>
-    {% endif %}
 
     {% include "two_factor/_wizard_actions.html" %}
   </form>
+
+  {% block 'backup_tokens' %}
+    {% if backup_tokens %}
+       <hr/>
```
claudep, line 40: nitpick, but I don't think we need the xhtml syntax nowadays, I would simply write it `<hr>`.

two_factor/templates/two_factor/core/login.html
```diff
@@ -31,14 +31,23 @@ <h1>{% block title %}{% trans "Login" %}{% endblock %}</h1>
...
+       <small>
```
claudep, line 41: I'm not convinced about the `small` usage. I'd rather include the section in a `<div>` with a class name so site authors can easily style it through their CSS.

claudep (changes requested):

two_factor/templates/two_factor/core/login.html
PetrDlouhy: OK

two_factor/templates/two_factor/core/login.html
PetrDlouhy: OK.

claudep (approved):

claudep:
Thanks I'll let the PR open some days to let others some time to comment.

PR #547, opened 2022-09-26
Allow running e2e tests with Chromium

claudep:
When Google Chrome is not available for running Selenium tests, try to fallback to Chromium, if present.


peymanslh (approved):
LGTM!

PR #550, opened 2022-09-27
Increase template overridability

PetrDlouhy:
I have added few code blocks to `login.html` template. This reduces the need for code duplication in overriding templates.

This PR contains commit from #546 to eliminate need to code changes after that PR gets merged. I expect that #546 well be merged first.

## Description
Code blocks allows to override only part of given template.

## Motivation and Context
Reduce code duplication.

## How Has This Been Tested?
Tox, manually

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/templates/two_factor/core/login.html
+++ b/two_factor/templates/two_factor/core/login.html
@@ -19,7 +19,9 @@
       enter one of these backup tokens to login to your account.{% endblocktrans %}</p>
   {% endif %}
 
-  <form action="" method="post">{% csrf_token %}
+  <form action="" method="post">
+    {% block main_form_content %}
+    {% csrf_token %}
     {% include "two_factor/_wizard_forms.html" %}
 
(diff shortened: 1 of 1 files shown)


PetrDlouhy:
@claudep Can you please review this? I have added the form extra params in separate PR as you asked as well as blocks to increase the template overridability.

claudep:
It seems to me that there are a lots of blocks now, isn't this a bit too much? I guess that if you really need all those blocks, you'd rather write your own template. We could add blocks that match an obvious and frequent use case. That's just an overall feeling, and I'd appreciate other opinions on the patch.

PetrDlouhy:
@claudep Thank you for your comment. I guess you are right, I can achieve almost the same code reusage with only one additional block in the original template.
The main form content block enables me to reuse quite some block of code, and I expect it could be helpful also for others.

Would this be sufficient to get merged?

claudep (approved):
Looks good and simple!

Issue #145, opened 2016-03-27
`addstatictoken` from Django-OTP doesn't always work with BackupView etc.


moggers87:
This is partly a bug with Django-OTP. [See this PR](https://bitbucket.org/psagers/django-otp/pull-requests/18/expose-default-static-device-name-used-by/diff)

If the specified user has already generated backup tokens (and thus has a StaticDevice), then `addstatictoken` will behave as expected - the user gets a new backup token.

However, if the user didn't previously, the command will create a StaticDevice with the name "Backup Code". Django Two-Factor Authentication has the name "backup" hardcoded into all its views and so never sees the newly created token.

tl;dr: D2FA and Django-OTP have hardcoded different names for backup tokens.

moggers87:
Having thought about this some more, I'm wondering why D2FA views only look for a single StaticDevice - is there actually a use-case where a user would have backup and not-backup tokens?

psagers:
The name field of `django_otp.models.Device` is documented as "A human-readable name to help the user identify their devices." It's really not intended to identify specific devices in the database. If D2FA would like to manage specific devices, my recommendation would be to create a model that holds foreign keys to those devices.

Ideally, `addstatictoken` should probably fail if there's no `StaticDevice`, perhaps with a flag to enable the current auto-create behavior. This is really just meant to help with bootstrapping a development environment or something.

Either way, if D2FA wants to be in a position for `addstatictoken` to always do the right thing, that probably means ensuring that every user always has a `StaticDevice`, even if it has no tokens. Alternatively, it could provide its own management command for this kind of thing.

Bouke:
Good thoughts @psagers, that sounds like good improvements. Thanks for the advice on how to improve this package.

MarkusH:
@moggers87 
> Having thought about this some more, I'm wondering why D2FA views only look for a single StaticDevice - is there actually a use-case where a user would have backup and not-backup tokens?

I'm not saying it's a good idea and I'm not suggesting we should support it, but a lot a banks still provide printed TAN lists to use for e.g. online banking. These codes are static 2FA codes but are not backup codes in case you loose a device, afaict.

moggers87:
Closed via #560

PR #560, opened 2022-10-14
issue #145 fix to use any static backup device instead of strictly named 

nnseva:
## Description
Now any static OTP device can be used as a backup OTP device instead of one strictly having the name `'backup'`.

## Motivation and Context
The `django_otp.plugins.otp_static` package has a management command `addstatictoken` to create a special backup static token that could be used for the OTP backup process.

As described in issue #145 the `addstatictoken` creates a static device named differently from the static device default name used in the `two_factor` package, if no other backup devices are present for the user yet. Because of that, no backup button is shown while login in when the backup code has been created by the `addstatictoken` management command.

## How Has This Been Tested?
Two unit tests have been created altering backup static OTP device names in existing ones.

You can check working the patch yourself. Create your own project using the package, and look at the login process. An unpatched package doesn't show the backup button if the backup code has been created by the `addstatictoken` management command, while the patched one does.

## Types of changes

## Checklist:


two_factor/views/core.py
```diff
@@ -642,7 +635,10 @@ class BackupTokensView(FormView):
     number_of_tokens = 10
 
     def get_device(self):
-        return self.request.user.staticdevice_set.get_or_create(name='backup')[0]
```
moggers87, line 659: `get_or_create` accepts a defaults parameter: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#django.db.models.query.QuerySet.get_or_create

moggers87 (changes requested):
One minor change required, otherwise looks good :+1:

nnseva:
@moggers87 done

moggers87 (approved):
LGTM :+1:

PR #582, opened 2023-01-15
Exclude webdriver-manager 3.8.5 when installing e2e requirements

moggers87:
That version appears to break things

## Description
Excludes webdriver-manager 3.8.5 from being installed

## Motivation and Context

Tests have been broken ever since webdriver-manager 3.8.5 was released. Nothing appears to be broken in django-two-factor-auth. Hopefully this issue will be fixed in a future version of webdriver-manager which is why I am only excluding that one version.

## How Has This Been Tested?

Ran tests webauthn tests under Python 3.11 locally and saw that tests passed

## Types of changes

## Checklist:

### DIFF
--- a/requirements_e2e.txt
+++ b/requirements_e2e.txt
@@ -1,3 +1,3 @@
 # test with selenium
 selenium~=4.4.3
-webdriver-manager~=3.8.3
+webdriver-manager~=3.8.3,!=3.8.5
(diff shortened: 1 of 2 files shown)


claudep (approved):
Awesome! Thanks a lot for this finding (@jpaniagualaconich, good news!)

jpaniagualaconich:
great news indeed! thanks a lot!

PR #597, opened 2023-02-20
Squash two_factor migrations

claudep:


claudep:
Was a bit strange to squash the phone plugin migration while there was only one migration, but it was necessary because the initial migration was a state-only migration.

peymanslh (approved):

stryker-sk:
Thanks for this!

PR #598, opened 2023-02-22
Upgrade isort and flake8 in pre-commit config

peymanslh:
Fix failing of pre-commit ci

## Description
I think the failure was because of an issue in `isort` https://github.com/PyCQA/isort/pull/2078
So I upgrade it to `5.12.0`
I also upgraded the flake8 to  `6.0.0`


claudep (approved):
:+1:

PR #599, opened 2023-02-22
Update Django and Python version in doc and README

peymanslh:
Update Python and Django supported versions based on [CHANGELOG.md](https://github.com/jazzband/django-two-factor-auth/blob/master/CHANGELOG.md#1150) in documentation and README


claudep (approved):
Good catch!

Issue #601, opened 2023-02-22
Remove salted_hmac_sha256


sevdog:
Since now the minimun supported django version is 3.2
https://github.com/jazzband/django-two-factor-auth/blob/52b78a72fbfdeb356cadbc8faf23ac65a2141cc7/setup.py#L15

It does not make sense to keep this backported method and not use django `salted_hmac` instead.

https://github.com/jazzband/django-two-factor-auth/blob/52b78a72fbfdeb356cadbc8faf23ac65a2141cc7/two_factor/views/utils.py#L303-L310

claudep:
Good catch, do you plan to suggest a pull request?

sevdog:
@claudep I can provide a PR with this with ease.

PR #602, opened 2023-02-23
Remove backported salted_hmac use django provided one instead

sevdog:
## Description
Removes  `salted_hmac_256` from the codebase, replacing with `django.utils.crypto.salted_hmac`.

Also a small tweak: using `tuple` instead of `list` when there is no need to modify a sequence which is only needed to match an array-like API.

## Motivation and Context
As stated in #601: this package now has a minimum requirements for `django>=3.2`, thus the backported method `salted_hmac_256` is not required anymore as we can rely on `salted_hmac` from django.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue), refactor

## Checklist:

### DIFF
--- a/two_factor/views/utils.py
+++ b/two_factor/views/utils.py
@@ -1,6 +1,4 @@
 import base64
-import hashlib
-import hmac
 import logging
 import time
 
@@ -8,6 +6,7 @@ from django.conf import settings
 from django.contrib.auth import load_backend
 from django.core.exceptions import SuspiciousOperation
 from django.core.signing import BadSignature, SignatureExpired
+from django.utils.crypto import salted_hmac
 from django.utils.decorators import method_decorator
 from django.utils.encoding import force_bytes
 from django.utils.translation import gettext as _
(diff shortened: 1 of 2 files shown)


two_factor/views/utils.py
```diff
@@ -291,20 +290,10 @@ def validate_remember_device_cookie(cookie, user, otp_device_id):
 
 
 def hash_remember_device_cookie_key(otp_device_id):
-    return str(base64.b64encode(force_bytes(otp_device_id)))
+    return force_str(base64.b64encode(force_bytes(otp_device_id)))
```
claudep, line 293: Isn't this another issue you reported? If yes, please create a separate PR for it.

sevdog: Yes, if it is fine I can address it in this PR (also with the code to keep backward compatibility).

two_factor/views/utils.py
```diff
@@ -104,7 +103,7 @@ def get_prev_step(self, step=None):
         if step is None:
             step = self.steps.current
         form_list = self.get_form_list()
-        keys = list(form_list.keys())
+        keys = tuple(form_list.keys())
```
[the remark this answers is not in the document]
sevdog: I understad, this is just a little improvement that I was feeling bad to open a dedicated PR for.

I can revert this.

claudep (approved):
Thanks!

claudep:
I think a last improvement could be to reference the issue in the commit message (`Fixes #601` or something like that) so it will auto-close.

sevdog:
Done.

PR #603, opened 2023-02-23
Updated Transifex config file to new format

claudep:
The new Transifex cli tool updated its config format (see https://github.com/transifex/cli).


peymanslh (approved):
LGTM!

Issue #436, opened 2021-10-29
Running migrations shows missing expected migrations due to DEFAULT_AUTO_FIELD changes in Django


vaughnkoch:
This looks like the same issue as https://github.com/django-otp/django-otp/issues/93. 

## Current Behavior

```
$ python manage.py migrate
Operations to perform:
  Apply all migrations: admin, auth, authtoken, contenttypes, otp_static, otp_totp, sessions, two_factor, users
Running migrations:
  No migrations to apply.
  Your models in app(s): 'otp_static', 'otp_totp', 'two_factor' have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

## Expected Behavior
Running migrate should not show an error.

## Possible Solution

The simplest solution seems to be to specify an explicit primary key for this model.  A workaround is to set DEFAULT_AUTO_FIELD in settings.py so Django doesn't complain.

`DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'`

See more info at https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys.

## Steps to Reproduce

- Install Django 3.2.8
- Install django-two-factor-auth  1.13.1
- `python manage.py migrate`

## Your Environment
Django 3.2.8
django-two-factor-auth  1.13.1
django-otp 1.1.1

moggers87:
Specifying an explicit PK would be the best way to go IMO.

peymanslh:
I think we could close this issue, as far as Django adds `DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'` as a default in new projects `settings.py` and a new installation of `django-two-factor-auth` doesn't have this problem anymore.

claudep:
I don't think we can depend on projects setting `AutoField` as default. I'd suggest to follow django-otp approach in https://github.com/django-otp/django-otp/commit/7ef5b25c8a715 (setting `default_auto_field` in default app config).

peymanslh:
Sure, I work on this issue and update my PR based on your suggestion.

PR #608, opened 2023-02-26
Add default_auto_field in apps config

peymanslh:
~~Add DEFAULT_AUTO_FIELD in the examples setting.~~
Add default_auto_field in apps config.

Closes: #436

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+### Fixed
+- Set `default_auto_field` to `AutoField` in apps config that have models,
+  so no migrations are generated for projects defaulting to `BigAutoField` (#436).
+
 ## 1.15.1
 ### Fixed
 - Missing plugin templates (#583).
--- a/two_factor/plugins/phonenumber/apps.py
+++ b/two_factor/plugins/phonenumber/apps.py
@@ -8,6 +8,7 @@ from two_factor.plugins.registry import registry
 class TwoFactorPhoneNumberConfig(AppConfig):
     name = 'two_factor.plugins.phonenumber'
     verbose_name = "Django Two Factor Authentication – Phone Method"
+    default_auto_field = "django.db.models.AutoField"
 
     def ready(self):
         register_methods(self, None, None)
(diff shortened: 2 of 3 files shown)


claudep:
I'd suggest https://github.com/jazzband/django-two-factor-auth/issues/436#issuecomment-1445363150 instead.

claudep (changes requested):
Thanks! However I think we should only add it to apps that have models.

claudep:
You may also include a note in the Changelog (in a new Unreleased section).

two_factor/apps.py
```diff
@@ -5,6 +5,7 @@
 class TwoFactorConfig(AppConfig):
     name = 'two_factor'
     verbose_name = "Django Two Factor Authentication"
+    default_auto_field = "django.db.models.AutoField"
```
claudep, line 8: Is this needed? I think the `two_factor` app has no longer models.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+### Fixed
+- Add `default_auto_field` in apps config.
```
claudep, line 5: I would try to explain the change in more details (defaulting to AutoField...)

two_factor/apps.py
peymanslh: I saw that there are no models in the `two_factor` app but there was a migration directory in it and it was confusing for me.

CHANGELOG.md
peymanslh: Fixed it.

two_factor/apps.py
peymanslh: Fixed it.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+### Fixed
+- Set `default_auto_field` to `AutoField` in apps config that have models for Django>=3.2.
```
claudep, line 5: I would write: `Set default_auto_field to AutoField in apps config that have models, so no migrations are generated for projects defaulting to BigAutoField`. What do you think?

claudep: Oh and I would also mention the ticket number at the end of the note `(#436)`

peymanslh:
@claudep 
Thanks for your review and suggestion

claudep (approved):
Thanks!

PR #610, opened 2023-03-02
drop unneeded unique on public_key (#594)

### LINKED ISSUE #594: Migration for webauthn fails due to unique index on LONGTEXT not supported by MySQL/MariaDB
## Expected Behavior

The Migration plugins/webauthn/migrations/0001_initial.py should run fine in MySQL

## Current Behavior

The migration plugins/webauthn/migrations/0001_initial.py fails with an error complaining you can't create an index on BLOB./TEXT columns without specifying a key length.

## Possible Solution

- Use a prefix-index => bad, because it only guarantees the prefix is unique, not the whole key
- Verify uniqueness in python-code => Bad, because it's a performance bottleneck
- Drop the uniqueness requirement => Probably the best idea, the public key attestation in WebAuthN devices should be unique anyways.

1. Install django-two-factor-auth[webauthn]
2. Run ./manage.py migrate

## Context

This is a problem in MySQL/MariaDB for the most part because they don't support arbitrary length indexes on TEXT fields at all.
I've found reports of PostgreSQL exposing similar problems if the stored value is bigger than the pagesize of the Database Engine Index buffer, but couldn't verify that myself.

All in All it seems like a bad idea to Index a textfield that can store up to 1024 bytes (as of WebAuthN Level 5) or even infinite amounts of data (up until WebAuthN Level 4)

## Your Environment

* django-two-factor-auth version: 1.15
* Database: MariaDB 10.3

jpaniagualaconich:
## Description

Drops the unique constraint on `Webauthn.public_key`.

Modifies two_factor_webauthn migration 0001 so public_key doesn't start with a unique constraint, otherwise it'll fail if the db is MariaDB.

Adds 0002 to drop the unique constraint. It'll still run fine even if there is no unique constraint to drop, resulting in a no-op in that case.

## Motivation and Context

Fixes #594 .
## Types of changes

## Checklist:


two_factor/plugins/webauthn/migrations/0002_alter_webauthndevice_public_key.py
```diff
@@ -0,0 +1,18 @@
...
+        ("two_factor_webauthn", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="webauthndevice",
+            name="public_key",
+            field=models.TextField(),
+        ),
+    ]
```
claudep, line 18: As Django computes the SQL operations to do by comparing the migrations (and not with the real state of the database), this second migration will be a noop (as `python manage.py sqlmigrate two_factor_webauthn 0002` will tell you) and therefore is not useful (and the unique constraint will stay for existing installations).

jpaniagualaconich: ah, lol, I see. I was completely fooled by it. Indeed, I thought the no-op came from comparing the migration to DB state.

Ugh, will try something different that leaves everything undisturbed unless it's exactly the case described in #594 and get back to you.

jpaniagualaconich:
OK, ready for another round.

Admittedly a not-super-elegant solution but I like its simplicity.

claudep (changes requested):
Thanks for trying again, but adding a new setting just for that issue is not an option in my opinion. I would try to subclass `CreateModel` in the initial migration, and alter the model fields in one of its method, `state_forwards` or `database_forwards`, depending on the database connection vendor.

jpaniagualaconich:
Been there, tried that. Tinkering with `CreateModel`, `state_forwards` and `database_forwards` means too much wading through the entrails of the migration infrastructure.

Also, I didn't want to depend on an undocumented `vendor` attribute.

Anyway, let me propose a somewhat different solution that leaves `CreateModel` and `AlterField` alone and instead modifies `Migration.apply` and `unapply` directly and is perhaps less hacky, at least in my opinion.

claudep (commented):
I like that! Now maybe a dumb question, if we solved the initial migration with index creation on MySQL, can we simply keep that and avoid the second migration?

claudep:
#616 should solve the test failures.

jpaniagualaconich:
I'm afraid I don't fully understand your question. Avoid the second migration in which sense?

Applying 0002 for MySQL results in a no-op, so for practical purposes we are avoiding the second migration already.

claudep:
My (maybe stupid) question was that if we fixed the MySQL issue, can we keep the unique index?

jpaniagualaconich:
We want model state to be the same in all cases, MySQL or not, after 0002 is applied so a future 0003 migration won't need to do acrobatics like 0001 and 0002 are forced to do now.

When the executor builds the state of the model it uses `Migration.mutate_state` which doesn't get `schema_editor`, meaning that what we can't change things for MySQL since we don't know the database.

As a result, the non-MySQL versions of 0001 and 0002 are used. The consequence of this is that the state of `WebauthnDevice` is wrong for MySQL after `mutate_state` for 0001, getting fixed only after `mutate_state` for 0002.

In any case, having that pesky `unique=True` was totally unneeded in the first place.

claudep:
That makes a lot of sense. Can you please rebase your PR now? Tests should hopefully pass.

jpaniagualaconich:
deprecated codecov gave up the ghost?

claudep:
Looks so... I will not regret it.

claudep:
Looks like #619 fix this failure.

claudep:
Could you please rebase on master ? #619 was merged.

jpaniagualaconich:
done. Looks ready for the merge.

claudep (approved):
Thanks for your efforts!

Issue #587, opened 2023-01-25
On a clean install, migrations are not in sync with code


jessamynsmith:
## Expected Behavior
I would expect that when creating a new project, installing django-two-factor-auth, and running migrate, that no further migrations would need to be made for the two factor app.

## Current Behavior
When installing django-two-factor-auth and running makemigrations, it is generating a migration for the two_factor app:

```
(venv) % python manage.py migrate      
Operations to perform:
  Apply all migrations: admin, auth, contenttypes, my_app, otp_email, otp_static, otp_totp, sessions, two_factor
Running migrations:
  Applying otp_email.0001_initial... OK
  Applying otp_email.0002_sidechanneldevice_email... OK
  Applying otp_email.0003_emaildevice_email... OK
  Applying otp_email.0004_throttling... OK
  Applying otp_static.0001_initial... OK
  Applying otp_static.0002_throttling... OK
  Applying otp_totp.0001_initial... OK
  Applying otp_totp.0002_auto_20190420_0723... OK
  Applying two_factor.0001_initial... OK
  Applying two_factor.0002_auto_20150110_0810... OK
  Applying two_factor.0003_auto_20150817_1733... OK
  Applying two_factor.0004_auto_20160205_1827... OK
  Applying two_factor.0005_auto_20160224_0450... OK
  Applying two_factor.0006_phonedevice_key_default... OK
  Applying two_factor.0007_auto_20201201_1019... OK
  Applying two_factor.0008_delete_phonedevice... OK
(venv) % python manage.py makemigrations
Migrations for 'two_factor':
  venv/lib/python3.11/site-packages/two_factor/migrations/0009_initial.py
    - Create model PhoneDevice
```

## Possible Solution
If the issue is that a migration is missing, generate and commit it. If the problem is that a model is in the wrong app, perhaps remove it from that app. Note: this issue was introduced in version 1.15.0; if I install 1.14.0 there are no migration issues.

## Steps to Reproduce (for bugs)
1. Follow the installation instructions for django-two-factor auth
2. Run: ``` python manage.py makemigrations```

## Context
This issue can be avoided by only ever generating migrations for my own apps, but that is a bit tedious.

## Your Environment
* Browser and version: N/A
* Python version: 3.11.0
* Django version: 4.1.5
* django-otp version: 1.1.4
* django-two-factor-auth version: 1.15.0
* Link to your project: https://github.com/jessamynsmith/django_4_example

rikva:
[...]

PR #596, opened 2023-02-19
Fixes #587 - Avoid unneeded migration for PhoneDevice

claudep:
Also refs #469.

PR #620, opened 2023-04-24
Refs #587 - Avoid importing phonenumber stuff in main migrations

claudep:


mgax:
FWIW, the migration file `two_factor/migrations/0009_initial.py` is still being generated, because of the URLs in `two_factor/urls.py`. If I remove them too, no migration is generated any more.

claudep:
Sure, this is not sufficient to solve the migration issue. However I separated from #596 as this is IMHO not subject to any discussion, while the rest of the #596 PR might have some backwards compatibility issues.

peymanslh (approved):

claudep:
Thanks @peymanslh !

PR #631, opened 2023-06-21
Prepare the 1.15.3 release

claudep:


dekkers (approved):

lgarron:
Thanks for the release!

This is sufficient for installing from `pip`, but the project front page only lists 1.15.1 as the latest tagged release on GitHub. Do you plan to add newer releases there as well? https://github.com/jazzband/django-two-factor-auth/releases

claudep:
I'm not a fan of multiple locations for the same functionality. Do we need GitHub releases when we have PyPi? But agree that we should either keep updating it or clean it up altogether.

PR #635, opened 2023-07-03
Add json field to PhoneDevice model

NewGlad:
## Description
JSON field to PhoneDevice model to have an ability to store arbitary data

## Motivation and Context

In my project in addition to the existing field declared in the PhoneDevice model, I also want to store some additional data, such as request id, timestamp of last sms sent, and some more
I think that it is a very common desire, so looks like it is very convenient to have some field with arbitrary structure to have a way to store some additional data in the standard model

## Types of changes

## Checklist:


moggers87:
Not sure I see the need for this in django-two-factor-auth. You can get the same effect by using Django's multi table inheritance i.e. subclass `PhoneDevice` and use your subclass when you want to access the additional fields.

Closing this PR as it's over 6 months old, feel free to comment if you disagree :smiley_cat:

PR #637, opened 2023-07-11
Change login page to a variable setting

rgcarrasqueira:
the user wish to custom urls for two_factor and I think doing in this way could be more flexible

## Description

creating a variable to receive the default condition, and If customized it at settings receive my new values

## Motivation and Context
We cannot have hard code situations in a plugin

### DIFF
--- a/two_factor/views/mixins.py
+++ b/two_factor/views/mixins.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib.auth import REDIRECT_FIELD_NAME
 from django.contrib.auth.views import redirect_to_login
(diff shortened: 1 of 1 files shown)


moggers87:
Users can already set `login_view` to whatever value they need either by overriding it in a sublcass or via `as_view`: https://docs.djangoproject.com/en/5.1/topics/class-based-views/#usage-in-your-urlconf

E.g.:

```
class ExampleView(OTPRequiredMixin, TemplateView):
    login_url = settings.LOGIN_URL

# or

ExampleView.as_view(login_url=settings.LOGIN_URL)
```

PR #645, opened 2023-08-14
Fixes #634 - Avoid direct dependency on pydantic

### LINKED ISSUE #634: Please relax pydantic dependency
For the `webauthn` extra, there is a dependency on `pydantic` 1.x.

Pydantic has now released 2.0 with a lot of improvements, which I need in a project that also uses django-two-factor-auth.

I have analyzed the situation:

* `py_webauthn` (probably wrongly) allows all Pydantic versions, it ignores semantic versioning
* For pydantic 2.0, it looks like this is ok, as no features used by webauthn have broken in Pydantic 2.0
* django-two-factor-auth only imports an exception to catch it, which is also still in place

So, in order to allow newer Pydantic version, I propose one of the following:

## Remove the `pydantic` dependency

django-two-factor-auth itself does not need it; the exception class can be re-imported from `webauthn` instead

## Relax the dependency to `>=1.9.0,<3.0`

Allow pydantic 2.0 explicitly instead

claudep:
Thanks to Dominik George for the suggestion.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
     extras_require={
         'call': ['twilio>=6.0'],
         'sms': ['twilio>=6.0'],
-        'webauthn': ['webauthn>=1.6.0,<1.99'],
+        'webauthn': ['webauthn>=1.11.0,<1.99'],
         'yubikey': ['django-otp-yubikey'],
         'phonenumbers': ['phonenumbers>=7.0.9,<8.99'],
         'phonenumberslite': ['phonenumberslite>=7.0.9,<8.99'],
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
     extras_require={
         'call': ['twilio>=6.0'],
         'sms': ['twilio>=6.0'],
-        'webauthn': ['webauthn>=1.6.0,<1.99', 'pydantic>=1.9.0,<1.99'],
+        'webauthn': ['webauthn>=1.6.0,<1.99'],
         'yubikey': ['django-otp-yubikey'],
         'phonenumbers': ['phonenumbers>=7.0.9,<8.99'],
         'phonenumberslite': ['phonenumberslite>=7.0.9,<8.99'],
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -21,7 +21,7 @@
     extras_require={
         'call': ['twilio>=6.0'],
         'sms': ['twilio>=6.0'],
-        'webauthn': ['webauthn>=1.6.0,<1.99', 'pydantic>=1.9.0,<1.99'],
+        'webauthn': ['webauthn>=1.6.0,<1.99'],
```
jpaniagualaconich, line 24: shall we follow @Natureshadow 's suggestion "partially" and relax the dependency instead of dropping it?

It's the official path for the symbol until at least <3.0.

Re-importing from `webuthn.entrails.that.might.change` seems mildly more dangerous even if it lets us drop the direct dependency on pydantic.

claudep: Sorry, I don't understand why dropping the pydantic dependency is problematic, compared to other py-webauthn dependencies.

jpaniagualaconich: the difference between pydantic and other py_webauthn dependencies is that django-two-factor explicitly depends on ValidationError.

claudep: OK, now I understand your rationale. However, in my opinion, our test suite is enough to guard against a future potential removal of pydantic.ValidationError. I'm always for a minimal set of dependencies, when acceptable, as they might cause more harm than good.

claudep:
Looks like this PR is ready for merge now? @jpaniagualaconich, please give your approval if it's the case.

jpaniagualaconich (approved):

jpaniagualaconich:
@claudep  It's ready :+1:

claudep:
Thanks a lot for your contribution.

claudep:
However, you need to approve the patch formally so as I can merge it.

jpaniagualaconich (approved):

jpaniagualaconich:
@claudep I approved it twice now :thinking:

claudep:
Oh, but maybe you are not a Jazzband member? (https://jazzband.co/) You definitely deserve it! I encourage you to join.

jpaniagualaconich:
that explains things! :laughing: 

thanks! I'm a jazzband member now.

claudep:
Congrats :medal_sports:

PR #646, opened 2023-08-14
Drop support for EOL Python 3.7

claudep:


dekkers (approved):

claudep:
Thanks for merging! Generally, I found very little value having a merge commit, so I prefer a "rebase and merge" most of the time. But that's not a big deal...

PR #658, opened 2023-09-10
avoid multiple registrations of the same method

### LINKED ISSUE #657: same method can be registered multiple times
The plugin registry does not check if the method is already registered before registering it again.

## Expected Behavior
It should silently ignore the registration of a previously registered method.

## Current Behavior
It adds the method.

jpaniagualaconich:
Fixes #657 

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/plugins/registry.py
+++ b/two_factor/plugins/registry.py
@@ -77,6 +77,10 @@ class MethodRegistry:
         self.register(GeneratorMethod())
 
     def register(self, method):
+        for registered_method in self._methods:
+            if method.code == registered_method.code:
+                return
+
         self._methods.append(method)
 
     def unregister(self, code):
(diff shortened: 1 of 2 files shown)


claudep:
Once again, it would be great to add a small regression test.

jpaniagualaconich:
@claudep added `tests/test_registry.py`

two_factor/plugins/registry.py
```diff
@@ -77,6 +77,10 @@ def __init__(self):
         self.register(GeneratorMethod())
 
     def register(self, method):
+        for registered_method in self._methods:
+            if method.code == registered_method.code:
+                return   # Already registered, ignore.
```
claudep, line 82: Are you OK with the comment I added?

jpaniagualaconich: totally :+1:

claudep (approved):

Issue #639, opened 2023-07-18
Relaxing django-phonenumber-field version requirement?


apkallum:
Hello! I wanted to ask if there is a reason why django-phonenumber-field is pinned to below version 7? If there is no reason, could you update the version? I can open a little PR if needed.

claudep:
Sure, please check the Changelog of django-phonenumber-field to see if any change might impact django-two-factor-auth and then suggest a PR.

claudep:
@apkallum, gently ping about a PR for this issue

mlec1:
@claudep I made an MR to update `django-phonenumber-field` to `v7`. I didn't notice any breaking changes affecting this library. The unit tests seem to pass successfully

claudep:
Fixed by 2b4e8a5fe0b663812009ad3652863cee1d04604f

PR #664, opened 2023-09-10
feat(dependencies): Upgrade  django-phonenumber-field to v7

mlec1:
## Description
Upgrade django-phonenumber-field to v7

Refs:
- https://github.com/jazzband/django-two-factor-auth/issues/639

## Motivation and Context
Refs:
- https://github.com/jazzband/django-two-factor-auth/issues/639

## How Has This Been Tested?
Locally tested with the example app + run the unit tests successfully

## Types of changes

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         'Django>=3.2',
         'django_otp>=0.8.0',
         'qrcode>=4.0.0,<7.99',
-        'django-phonenumber-field>=1.1.0,<8',
+        'django-phonenumber-field<8',
         'django-formtools',
     ],
     extras_require={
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -15,7 +15,7 @@
         'Django>=3.2',
         'django_otp>=0.8.0',
         'qrcode>=4.0.0,<7.99',
-        'django-phonenumber-field>=1.1.0,<7',
+        'django-phonenumber-field>=1.1.0,<8',
```
claudep, line 18: Considering that version 1.1.0 was released in 2016, do you think we can drop the `>=1.1.0` part?

mlec1: Well for me, it would be fine. Which version would you like to target `>=2.0.0` or `>=3.0.0` or more

claudep: Simply don't provide a minimal version, unless it's required by some specific change.

claudep (approved):
Thanks!

Issue #647, opened 2023-08-15
Migration error when upgrading to 1.15.4


knyghty:
## Expected Behavior
Migrations work.

## Current Behavior
Migrations fail.

## Possible Solution
?

## Steps to Reproduce (for bugs)
Running migrate after upgrading to 1.15.4 from 1.15.3. Presumably needs the `phonenumber` plugin.

## Context
```pytb
Traceback (most recent call last):
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/graph.py", line 133, in remove_replaced_nodes
    replacement_node = self.node_map[replacement]
                       ~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: ('two_factor', '0001_squashed_0008_delete_phonedevice')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/./manage.py", line 31, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/base.py", line 106, in wrapper
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/core/management/commands/migrate.py", line 117, in handle
    executor = MigrationExecutor(connection, self.migration_progress_callback)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/django/db/migrations/executor.py", line 18, in __init__
    self.loader = MigrationLoader(self.connection)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #667, opened 2023-09-15
Delete two_factor/migrations/0001_squashed_0008_delete_phonedevice.py

hyperkai:
This file causes the error below so I deleted it:

> django.db.utils.OperationalError: no such table: two_factor_phonedevice

## Types of changes

## Checklist:


hyperkai:
I deleted `two_factor/migrations/0001_squashed_0008_delete_phonedevice.py` because the file causes the error below:

django.db.utils.OperationalError: no such table: two_factor_phonedevice

hyperkai:
I deleted two_factor/migrations/0001_squashed_0008_delete_phonedevice.py because the file causes the error below:

django.db.utils.OperationalError: no such table: two_factor_phonedevice

pacharanero:
Not sure if this is directly related but we do also seem to have an issue with unapplied migrations having recently implemented `two_factor` in our app. We have a step in our Docker Compose file which runs `python manage.py migrate` but for some reason we don't understand this one migration is not being applied.

We get a warning in the log console "You have 1 unapplied migration(s). Your project may not work properly until you apply the migrations for app(s): two_factor." and when entering the relevant Docker container and running `python manage.py migrate` again, it then applies the migration: `Applying two_factor.0001_squashed_0008_delete_phonedevice... OK` but we don't see why this should be necessary when all our other migrations get applied automatically during the provisioning of the Docker Compose setup.

mlec1:
@hyperkai I would suggest opening an issue instead. 

Which version of the package are you using ? The table you mention is not in the file you want to delete. It is in the `phonenumber` plugin

moggers87:
Closing this as I think this is a duplicate of #647

PR #672, opened 2023-10-17
Add plivo gateway.

wieczorek1990:
Added Plivo gateway for SMS.
## Description

I added a new module in gateways for sending SMS using the phone number module.

## Motivation and Context

Plivo is an alternative to Twilio.

## How Has This Been Tested?

Manually tested with use of the setup page and read on the SMS debug page of Plivo.

## Types of changes

## Checklist:

- [ ] My change needs translation update. (moved a translated block of code from templates)


claudep (changes requested):
Thanks for this code! This still needs documentation on how to configure this gateway.

Moreover, I think you cannot simply change the name of the sms template, as people having overridden the template will suddenly see the default template instead of the custom one. I would suggest adding a list of two templates in `render_to_string`, so the old path is still used if it exists in user projects.

wieczorek1990:
Will add different template.

W dniu wt., 17.10.2023 o 17:22 Claude Paroz ***@***.***>
napisał(a):

wieczorek1990:
@claudep I do not see a possibility to use two `template_name`'s in `render_to_string`.

wieczorek1990:
@claudep Added documentation.

wieczorek1990:
I do not like this solution. The parameter check only said template_name.

W dniu wt., 17.10.2023 o 17:58 Claude Paroz ***@***.***>
napisał(a):

wieczorek1990:
@claudep can we go with a separate template?

W dniu wt., 17.10.2023 o 19:13 Łukasz Wieczorek ***@***.***>
napisał(a):

PR #675, opened 2023-10-18
Force fetching translation files from transifex

SupraSummus:
In `make tx-pull` make sure we always fetch most recent transifex translation.

## Motivation and Context
The problem is that in some circumstances `make tx-pull` may ignore changes made in transifex. This is because `tx pull -a` command checks file modification timestamps. Adding `--force` ensures we fetch all files regardless of their remote and local modification date.

Line from a help message of `tx pull`:

```
    --force, -f                  Force the download of the translations files regardless of whether timestamps on the local computer are newer than those on the server (default: false)
```

I encountered the problem in the following scenario:

1. I submitted some translations to transifex.
2. I forked and cloned the repo to fetch transifex changes.
3. My changes were not fetched because local modification times was after transifex modification times.

## How Has This Been Tested?
Having installed transifex cli I successfully ran `make tx-pull`.

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ coverage:
 	coverage report --precision=2
 
 tx-pull:
-	tx pull -a
+	tx pull -a --force
 	cd two_factor; django-admin compilemessages
 	cd example; django-admin compilemessages
 


claudep:
I'm not in favour of using `--force` systematically. tx pulling is generally done by maintainers who have checkouts existing for some time.

SupraSummus:
I know this is a weak argument, but django does `-f` (equal to `--force`): https://github.com/django/django/blob/97b7970c6479ef414175f22a009240ed0df8ce46/scripts/manage_translations.py#L155-L174

Another aspect is a counterargument for your statement that "maintainers have checkouts existing for some time": even a long-existing clone may have a recent modification date on a `.po` file. For example because of running `./manage.py makemessages`.

I would recommend going with `--force`. The only downside is a bit slower execution and a bit more stuff transferred over the network, but the advantage is that the possibility of a mistake lowered tenfold or more.

claudep (approved):
OK, your arguments convinced me :smile:

emergencyusaarmyvaccation (commented):
Nice

PR #679, opened 2023-11-02
Add whatsapp method

### LINKED ISSUE #678: Feature: Enable Whatsapp as a method
This feature allows offering WhatsApp as aa valid method in the package for users to choose from. WhatsApp is more secure than SMS.

## Expected Behavior

If the Django project implements the Twilio gateway for SMS it can also quite easily offer WhatsApp through Twillio. This feature will work exactly the SMS feature.

## Current Behavior

The difference from the current behavior is that users will now be able to select WhatsApp in addition to SMS or Call.

## Possible Solution

Implement the new method by simply duplicating the SMS pattern and adjusting for whatsapp accordingly.

1. N/A
2.
3.
4.

## Context

WhatsApp is an excellent source of 2FA particularly in developing nations where phone service and reliable SMS can be an issue. WhatsApp is secure and fast and improves the user experience since the message comes to the user as opposed to having to pull up an authenticator app.

## Your Environment

* Browser and version: Chrome
* Python version: 3.11
* Django version: 4.1
* django-otp version: 1.1.3
* django-two-factor-auth version: 1.15.5
* Link to your project: n/a

ryaustin:
This change takes advantage of the available phone methods and Twilio gateway to allow the user to select WhatsApp as their token method.

## Description

In the same pattern as the SMS method, the new WhatsApp method is added along with some user-definable settings. The change also exposes the ability to set the size of the token input field which was a bit small by default. Appropriate documentation has also been added.

## Motivation and Context

Fixes #678. WhatsApp has become a major communication protocol globally due to its reach, speed, and security. It seems natural that a user can receive a token via WhatsApp. It is also more secure than SMS.

## How Has This Been Tested?

Using existing mock tests in the pattern of SMS. Sending a WhatsApp is akin to sending a SMS. The example app implementation also works as intended. The documentation was tested within a vanilla example app for corectness.

Screenshot 2023-11-02 at 10 29 01 AM
Screenshot 2023-11-02 at 10 29 14 AM
Screenshot 2023-11-02 at 10 29 44 AM

## Types of changes

## Checklist:

- [ x ] My code follows the code style of this project.
- [ x] My change requires a change to the documentation.
- [ x ] I have updated the documentation accordingly.
- [ x ] I have added tests to cover my changes.
- [ x ] All new and existing tests passed.


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.15.5
+current_version = 1.16
```
moggers87, line 2: Can you remove the commits that change the version string? This is something that's done at release time.

docs/conf.py
```diff
@@ -59,7 +59,7 @@
 #
 
 # The full version, including alpha/beta/rc tags.
-release = '1.15.5'
+release = '1.16'
```
moggers87, line 62: Can you remove the commits that change the version string? This is something that's done at release time.

setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='django-two-factor-auth',
-    version='1.15.5',
+    version='1.16',
```
moggers87, line 5: Can you remove the commits that change the version string? This is something that's done at release time.

two_factor/plugins/phonenumber/models.py
```diff
@@ -8,11 +8,13 @@
 from django_otp.util import hex_validator, random_hex
 from phonenumber_field.modelfields import PhoneNumberField
 
-from two_factor.gateways import make_call, send_sms
+from two_factor.gateways import make_call, send_sms, send_whatsapp
 
+WHATSAPP = 'wa'
```
moggers87, line 13: Not sure what this is for, it's only used in a few places. Also why are we using `"wa"` rather than `"whatsapp"`?

two_factor/forms.py
```diff
@@ -27,13 +28,13 @@ def __init__(self, **kwargs):
...
 
 class DeviceValidationForm(forms.Form):
-    token = forms.IntegerField(label=_("Token"), min_value=1, max_value=int('9' * totp_digits()))
-
+    token = forms.IntegerField(label=_("Code"), min_value=1, max_value=int('9' * totp_digits()))
     token.widget.attrs.update({'autofocus': 'autofocus',
                                'inputmode': 'numeric',
-                               'autocomplete': 'one-time-code'})
+                               'autocomplete': 'one-time-code',
+                               'style': f'width:{TWO_FACTOR_INPUT_WIDTH}px; height:{TWO_FACTOR_INPUT_HEIGHT}px;'})
```
moggers87, line 35: This won't work if a site is using CSP and disallows inline styles

two_factor/forms.py
```diff
@@ -11,11 +11,12 @@
 from .plugins.registry import registry
 from .utils import totp_digits
 
+TWO_FACTOR_INPUT_WIDTH = getattr(settings, 'TWO_FACTOR_INPUT_WIDTH', 100)
+TWO_FACTOR_INPUT_HEIGHT = getattr(settings, 'TWO_FACTOR_INPUT_HEIGHT', 30)
```
moggers87, line 15: It's not clear that we need this

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.15.6
+### Changed
```
moggers87, line 4: This should be 

```
## Unreleased
### Added
```

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+- Added WhatsApp as a token method through Twilio
+- Added the ability to modify the token input size
```
moggers87, line 6: I don't think modifying the token input size should be a part of this PR

moggers87 (changes requested):
It's a good start, but there's a lot of stuff that shouldn't be here.

two_factor/plugins/phonenumber/models.py
ryaustin: explicit is better. agreed.

two_factor/forms.py
ryaustin: You are correct, I will remove this functionality from this pr.

ryaustin:
Appreciate the feedback, i'll make the changes as soon as I can.

PR #680, opened 2023-11-15
Adding TOTP_ISSUER as settings value

yopiti:
## Description
Using the site name at this point doesn't always make sense. This gives you the possibility to set TOTP_ISSUER in the settings.py file and show this information in the TOTP app.

## Motivation and Context
Depending on the setup, using the sitename could lead to just showing a short name instead of the application name / url. 

## How Has This Been Tested?
Code testing

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/views/core.py
+++ b/two_factor/views/core.py
@@ -707,6 +707,8 @@ class QRGeneratorView(View):
     }
 
     def get_issuer(self):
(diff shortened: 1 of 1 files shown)


claudep (changes requested):
Might make sense, however this needs docs, tests and of course the test suite must pass.

PR #683, opened 2023-12-13
Only validate token against chosen device 2 (#473)

PetrDlouhy:
This is rebase of #521 to the current master branch.

Please follow the description at #521

### DIFF
--- a/tests/test_views_login.py
+++ b/tests/test_views_login.py
@@ -406,6 +406,94 @@ class LoginTest(UserMixin, TestCase):
         # Check that the signal was fired.
         mock_signal.assert_called_with(sender=mock.ANY, request=mock.ANY, user=user, device=device)
 
+    def test_totp_token_does_not_impact_backup_token(self):
+        """
+        Ensures that successfully authenticating with a TOTP token does not
+        inadvertently increase the throttling count of the backup token device.
+
+        Addresses issue #473, where correct TOTP token usage was unintentionally
+        affecting the throttling count of backup tokens, potentially leading to
+        their invalidation.
+        """
+        user = self.create_user()
+        backup_device = user.staticdevice_set.create(name='backup')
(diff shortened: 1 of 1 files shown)


PetrDlouhy:
The tests are not passing because of #684

two_factor/forms.py
```diff
@@ -115,6 +116,7 @@ class AuthenticationTokenForm(OTPAuthenticationFormMixin, forms.Form):
         'pattern': '[0-9]*',  # hint to show numeric keyboard for on-screen keyboards
         'autocomplete': 'one-time-code',
     })
+    device_id = forms.CharField(widget=forms.HiddenInput(), required=False)
```
claudep, line 119: I don't see why we need this `device_id` hidden field, if the user has no way to change it anyway. Couldn't we return `initial_device` from `_chosen_device`?

two_factor/forms.py
```diff
@@ -154,11 +161,19 @@ def __init__(self, user, initial_device, **kwargs):
...
+
     def _chosen_device(self, user):
-        device = super()._chosen_device(user)
-        if not device:
-            device = self.initial_device
-        return device
+        if self.device_cache:
+            return self.device_cache
+        else:
+            return self.initial_device
```
claudep, line 176: If we skip calling the parent method, we should at least redo the security check done in `django_otp` parent method.

PetrDlouhy:
@claudep Thank you for your comments. I had to investigate this much deeper and I found that the issue #473 might be already resolved by https://github.com/jazzband/django-two-factor-auth/commit/8deb380eb3cbb9b27f90a8e822e4951c31856515

The test from the original PR is passing in current master and I added one more test to ensure that the throttling will work correctly in the future.

Would you be OK with merging just those tests?

claudep:
Absolutely, and thanks a ton for these investigations :heart_eyes:

PR #691, opened 2024-01-06
change nav item logout link from get link to post button

k-brahma:
As this is the first time tring to contribute OSS, sorry in advance if any of my procedures are not appropriate.

navbar logout link changed to post form.

## Description

example/templates/_base.html fixed to:

```html
<li class="nav-item">
<form action="{% url 'logout' %}" method="post" style="display: inline;">
    {% csrf_token %}
    <button type="submit" class="nav-link" style="background: none; border: none; padding-top: 8px; margin: 0; text-decoration: none; cursor: pointer;">
        {% trans "Logout" %}
    </button>
</form>
</li>
```

## Motivation and Context

original link below causes 405 response.
```<a class="nav-link" href="{% url 'logout' %}">{% trans "Logout" %}</a>```

## How Has This Been Tested?

the test below actually ends with some errors but not differ from origin/master's result.
```python
python .\example\manage.py test --settings=tests.settings   
```

## Types of changes

## Checklist:

### DIFF
--- a/example/templates/_base.html
+++ b/example/templates/_base.html
@@ -43,7 +43,7 @@
               <li class="nav-item">
                 <form action="{% url 'logout' %}" method="post" style="display: inline;">
                     {% csrf_token %}
-                    <button type="submit" class="nav-link btn btn-link p-2 m-0 text-decoration-none">
+                    <button type="submit" class="nav-link btn btn-link pt-2 m-0 text-decoration-none">
                         {% trans "Logout" %}
                     </button>
                 </form>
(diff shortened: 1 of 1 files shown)


claudep (changes requested):
Thanks for your patch! Indeed, the logout link should be a POST link. However, as the template is bootstrap-style, could you please avoid direct "style=" and replace the styles by matching bootstrap classes (like `d-inline`, etc.)?

k-brahma:
Hi, I fixed it.
Think that 'class="nav-link btn btn-link pt-2 m-0 text-decoration-none"' enough.

I visually confirmed that the nav-link works correctly in both expanded and collapsed states.

k-brahma:
d-inline

So I launched local sever and checked if the class propery works.
Think it's enough.

PR #700, opened 2024-01-28
Add testing for Django 5.0

claudep:


requirements_dev.txt
```diff
@@ -15,7 +15,7 @@ django-user-sessions
 
 # Example app (WebAuthn)
 
-webauthn~=1.6.0
+webauthn~=1.11.0
```
sergei-maertens, line 18: I find the setup with some dependencies in `setup.py`, then again in `tox.ini` and apparently some more in requirements files quite confusing. I missed this file in #702

Wouldn't it make more sense to make use of the `extras` feature and recommend to install using `pip install -e .[webauthn]` for development?

claudep: I do agree that the current dependency handling is dispersed. Please propose your suggestion in a separate PR, that would be welcome!

sergei-maertens (approved):
Looks good to me, I have some minor remarks that aren't blocking.

requirements_dev.txt
sergei-maertens: I'll take a look if I have some more time this evening! Probably a good idea to first get a grip on the open PRs, so I'll take a look at those too since my jazzband membership allows you to merge them :smile:

tox.ini
```diff
@@ -1,23 +1,23 @@
 [tox]
-; Minimum version of Tox
-minversion = 1.8
 envlist =
     py{38,39,310,311}-dj{32,40,41,42}-{normal,yubikey,custom_user,webauthn}
```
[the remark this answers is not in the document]
claudep: We may remove 4.0 and 4.1 in the list, but I don't care that much, to be honest.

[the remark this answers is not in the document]
sergei-maertens: Yeah, if it ain't broken... As long as it doesn't sit in the way it doesn't matter either.

PR #702, opened 2024-02-02
:sparkles: Added support for py webauthn 2.0+

### LINKED ISSUE #701: What is needed to upgrade to webauthn 2.0.0+?
While investigating some conflicts in tooling in their Pydantic versions used, I noticed that webauthn 2.0.0 was released a couple weeks ago which dropped their dependency on Pydantic (which would solve my tooling issue). However, django-two-factor-auth pins webauthn on `<1.99`.

Does anyone have an idea what's needed to support webauthn 2.0.0 and could they provide some guidance? I can probably allocate some $dayjob time as it would make our setup/needs simpler to achieve.

## Expected Behavior

Use django-two-factor-auth with webauthn 2.0.0

## Current Behavior

Only v1 of webauthn is supported.

## Possible Solution

See if/how the breaking changes affect the usage in django-two-factor-auth and if necessary:

* add a compatibility layer or 
* require webauthn v2.0.0+

The latter option should be safe for project that use Pydantic themselves, they will now only need to explicitly declare their dependency.

n/a

## Context

I'm trying to use the `bump-my-version` tool and django-two-factor-auth[webauthn] in the same virtualenv, but they have incompatible pydantic requirements.

As a workaround, I can install/run bump-my-version from a different venv.

## Your Environment

* Browser and version: Brave Version 1.62.153 Chromium: 121.0.6167.85 (Official Build) (64-bit)
* Python version: 3.10, 3.11, 3.12
* Django version: 4.2
* django-otp version: 1.3.0
* django-two-factor-auth version: 1.15.5
* Link to your project: https://github.com/maykinmedia/maykin-2fa

sergei-maertens:
## Description
Closes #701 

WebAuthn 2.0 refactored pydantic usage out of the codebase.

Tests were passing locally with both 1.x and 2.y, but for simplicity's sake, the minimum version is now set to 2.0 so that no compat layer is required.

There doesn't seem to have been a real reason to catch pydantic validation errors - nothing was documented in 96bbd1af and the model does not appear to perform additional validations (see
https://github.com/duo-labs/py_webauthn/blob/v1.11.1/webauthn/authentication/verify_authentication_response.py ), but possibly the base model did. Either way, it was not a tested execution branch.

## Motivation and Context

Drops dependency on Pydantic which may cause version conflicts with other tools.

## How Has This Been Tested?

`tox -e py311-dj42-webauthn` and having the Github CI pipeline. There shouldn't be any functional changes.

n/a

## Types of changes

## Checklist:

- `n/a` I have added tests to cover my changes.
- [x] All new and existing tests passed. (waiting for CI to run the rest)

### DIFF
--- a/two_factor/plugins/webauthn/forms.py
+++ b/two_factor/plugins/webauthn/forms.py
@@ -6,9 +6,9 @@ from django.urls import reverse_lazy
 from django.utils import timezone
 from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
-from pydantic import ValidationError as PydanticValidationError
 from webauthn.helpers.exceptions import (
-    InvalidAuthenticationResponse, InvalidRegistrationResponse,
+    InvalidAuthenticationResponse, InvalidJSONStructure,
+    InvalidRegistrationResponse,
 )
 from webauthn.helpers.parse_authentication_credential_json import (
     parse_authentication_credential_json,
@@ -91,7 +91,7 @@ class WebauthnAuthenticationTokenForm(WebauthnEntitiesFormMixin, AuthenticationT
 
             new_sign_count = verify_authentication_response(
                 device.public_key, device.sign_count, self.webauthn_rp, self.webauthn_origin, challenge, token)
-        except (PydanticValidationError, WebauthnDevice.DoesNotExist, InvalidAuthenticationResponse) as exc:
+        except (InvalidJSONStructure, WebauthnDevice.DoesNotExist, InvalidAuthenticationResponse) as exc:
             raise forms.ValidationError(_('Entered token is not valid.'), code='invalid_token') from exc
 
         device.sign_count = new_sign_count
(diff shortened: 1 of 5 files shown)


sergei-maertens:
I will investigate the issues on Python < 3.11!

two_factor/plugins/webauthn/forms.py
```diff
@@ -91,7 +90,7 @@ def _verify_token(self, user, token, device=None):
 
             new_sign_count = verify_authentication_response(
                 device.public_key, device.sign_count, self.webauthn_rp, self.webauthn_origin, challenge, token)
-        except (PydanticValidationError, WebauthnDevice.DoesNotExist, InvalidAuthenticationResponse) as exc:
+        except (WebauthnDevice.DoesNotExist, InvalidAuthenticationResponse) as exc:
```
claudep, line 93: Looks like you have to also catch `webauthn.helpers.exceptions.InvalidJSONStructure` here.

sergei-maertens: you're right, something was wrong in my `.tox` dependencies and the tests failed locally after nuking that :)

claudep:
May I ask you to approve my recent PRs, then you can rebase yours on top?

claudep (approved):
Thanks!

PR #705, opened 2024-02-03
Prepare 1.16.0 release

claudep:


claudep:
@sergei-maertens Do you think we should rather switch to `1.16.0`?

sergei-maertens:
I think 1.16.0 is more appropriate - the webauthn minimal version bump is a rather big change. I think it's also good to mention that users can probably uninstall pydantic if no other dependencies or their own project uses it.

With tooling like `pip-tools` this is managed automatically, but there are plently of people still just freezing their env into a requirements file and they may not need pydantic anymore.

claudep:
> but there are plently of people still just freezing their env into a requirements file

/me whistles innocently...

sergei-maertens (approved):

claudep:
Thanks a lot @sergei-maertens for your help today!

sergei-maertens:
You're welcome! Feel free to ping me if you're in need of reviewa again

PR #706, opened 2024-02-03
:hammer: Re-organize dependencies for local dev and CI

sergei-maertens:
The 'hard' dependencies are now specified as extra groups in the setup.py package information. tox.ini points to these extras to install the required dependencies, and their supported version ranges.

The dev requirements file is now documented in the contributing guidelines, and makes use of the extras to pull in the optional dependencies for a full-blown local installation. The remaining dependencies (such as sphinx) are not used in tox and are still in the requirements_dev.txt file.

The linting dependencies (with upper bound versions) are now also specified as an extra group to prevent mismatching versions in tox env and local development envs.

## Motivation and Context

`pip install -r requirements_dev.txt` lead to different versions of isort/flak8 being installed in local env and CI pipeline. While it
didn't cause problems yet, these are annoying to track down.

Additionally, when touching minimum required versions, it's hard to keep track of where the version numbers are all specified
and can lead to false negatives in the tests if there are things broken because one specification overrides another.

## How Has This Been Tested?

* Clean virtualenv
* `pip install -r requirements_dev.txt`
* `make example-webauthn`
* `tox`

n/a

## Types of changes
- [x] Project maintenance

## Checklist:

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -15,6 +15,16 @@ Contribute
 * Send a pull request with your changes.
 * Provide a translation using Transifex_.
 
+Local installation
+------------------
+
+Install the development dependencies, which also installs the package in editable mode
+for local development and additional development tools.
+
+.. code-block:: console
+
+    pip install -r requirements_dev.txt
+
 Running tests
 -------------
 This project aims for full code-coverage, this means that your code should be
--- a/docs/requirements.rst
+++ b/docs/requirements.rst
@@ -3,11 +3,12 @@ Requirements
 
 Django
 ------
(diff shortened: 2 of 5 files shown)


claudep (approved):
Great, thanks a lot! Sooner or later, we should also migrate to `pyproject.toml` instead of `setup.py`.

sergei-maertens:
yeah, I wanted to mention that earlier but it involves some changes in build tooling too. we recently updated our own starter template to it, maybe @viicos wants to do this one :wink:

Viicos:
Sure thing, always happy to migrate projects to the new PEP 517/8

PR #707, opened 2024-02-04
Replace `setup.py` by `pyproject.toml`

Viicos:
## Types of changes

## Checklist:


pyproject.toml
```diff
@@ -0,0 +1,99 @@
...
+name = "django-two-factor-auth"
+version = "1.16.0"
+description = "Complete Two-Factor Authentication for Django"
+readme = "README.rst"
+authors = [
+  {name = "Bouke Haarsma", email = "bouke@haarsma.eu"},
+]
+maintainers = [
+  {name = "Claude Paroz", email = "claude@2xlibre.net"},
+]
```
Viicos, line 15: More maintainers can be added.

claudep: I would be more than happy to see more names here. 🆘

setup.cfg
```diff
@@ -1,12 +1,3 @@
 [flake8]
 exclude = build,dist,docs,.tox,./two_factor/views/__init__.py
 max-line-length = 119
```
claudep, line 3: Do you think it's worth keeping `setup.cfg` for flake8 (or rename .flake8)? Anyway, I guess most projects turn to ruff these days.

claudep (commented):
Many thanks for your help! I'll let the PR open for some days, in the case more maintainers would like to be mentioned.

setup.cfg
Viicos: We could switch to Ruff to simplify tooling (removes isort and flake8 and add ruff format as well)?

claudep: I'm generally not a fan of black-like formatting tool. For me, some choices of those tools do not provide visual advantages and are mostly there to ease automatic formatting. Let's keep human touches in code formatting :smile:

Viicos: Alright as you wish :+1: So ruff for isort and flake8 or we keep flake8 but with a `.flake8` config file instead?

claudep: +1 for ruff for isort and flake8

pyproject.toml
```diff
@@ -0,0 +1,114 @@
...
+line-length = 119
+target-version = "py38"
+extend-exclude = ["docs"]
+
+[tool.ruff.lint]
+select = [
+    "F",  # Pyflakes
+    "E",  # pycodestyle (Error)
+    "W",  # pycodestyle (Warning)
+    # "I",  # isort (waiting for https://github.com/astral-sh/ruff/issues/2600)
```
Viicos, line 83: Let me know if you'd like to use isort from Ruff even if it would require some formatting changes to imports

claudep: If the changes are not too much invasive, I think it's fine.

tox.ini
```diff
@@ -60,10 +60,10 @@ commands =
     coverage run {env:COVERAGE_OPTIONS:} {envbindir}/django-admin test -v 2 --pythonpath=./ --settings=tests.settings
     coverage report
 
-[testenv:flake8]
+[testenv:ruff]
 basepython = python3
 extras = linting
-commands = flake8 example tests two_factor
+commands = ruff
```
dekkers, line 66: I think this will fix the failing github action:

```suggestion
commands = ruff .
```

pyproject.toml
Viicos: Yeah the thing is it was putting each imported object on a single line :/ Is it fine still? If the diff is too much we can add an entry to ignore the commit for blames

pyproject.toml
moggers87: I know I'm not exactly active at the moment, but I'd be happy to have my name here :+1:

Viicos:
Thanks for finishing up, was lacking some time

PR #715, opened 2024-03-21
Show/buttons by authentication

LomNtetha:
## Description

This is a new feature for wizard templates. It adds a new sign-in button to the wizard actions and includes authentication for users. The sign-in button will be displayed only when the user is not authenticated. Otherwise, it will display other buttons. This feature ensures that the sign-in button is hidden when the user is logged in, without interfering with the functionality of the templates

## Motivation and Context
This feature helps to remove the cancel, back, and previous buttons that are shown on the sign-in page when the user is not authenticated.

## How Has This Been Tested?
New tests have been created for this feature, and they pass successfully. The two test functions are named test_render_template_authenticated and test_render_template_not_authenticated.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_wizard.py
+++ b/tests/test_wizard.py
@@ -1,0 +1,25 @@
+from django.template.loader import render_to_string
+from django.test import SimpleTestCase
+
+class TestWizardActionsTemplate(SimpleTestCase):
+    def test_render_template_authenticated(self):
+        template_name = 'two_factor/_wizard_actions.html' 
+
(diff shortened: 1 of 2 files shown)


moggers87:
I'm not sure I understand the motivation for this change. If I understand what you've done right then an unauthenicated user would see "sign-in" rather than "next" on every step of a wizard, which would be a little confusing if there was more than one step IMO

PR #732, opened 2024-08-01
Remove bump2version dependency

claudep:
Keeping a (mostly unmaintained) dependency just to change a version number in two files looks a bit unnecessary to me.


moggers87 (approved):
:+1:

Issue #751, opened 2025-03-25
Email Device should be confirmed False as default


MAkcanca:
While setting up the OTP Email, even though we don't confirm the sent code and just abandon/cancel the flow, it still saves the email as confirmed device, causing user to have OTP/2FA enabled even though they have no valid devices.
We should set the Email device confirmed property to False as default. Need to also discuss possible side effects.

moggers87:
Closed via #752

Issue #562, opened 2022-10-20
Email Method, sends the email, then if i abandon the page view, it has enabled two factor authentication but I haven't completed the validation step


seen-it:
On the EmailMethod the validation step creates an EmailDevice (django-otp) which if I abandon the page view without entering my token seems to enable two factor authentication on that account, in a strange state too where i cannot disable the enabled 2fa on the account or finish the original setup for the account.

Probably should be able to finish an abandoned setup process if the user fails to enter the first validation token
 
The current behaviour is that if i select the email method, the web browser takes me straight to the validation step in which the user enters the token sent to the users email address, the user recieves that token but then abandons the session or closes the web page.
Then if the user attempts to view the profile screen to rerun the not completed 2fa setup it shows the account as enabled, in which if the user tries to disable there account it just takes them back to to LOGIN_REDIRECT_URL

seems like the issue stems from here
https://github.com/jazzband/django-two-factor-auth/blob/master/two_factor/views/core.py#L514
https://github.com/django-otp/django-otp/blob/master/src/django_otp/plugins/otp_email/models.py#L59

* Python version: 3.9.12
* Django version: 3.2.15
* django-otp version: 1.1.3
* django-two-factor-auth version: 1.14.0

inellas:
I have experienced the same problem. Do we have any update regarding this?

It seems quite important, as it may lead to a user being locked out of their account under certain circumstances:
A user wants to enable two factor authentication by email and enters a wrong email address by mistake.
When clicking next step, they realize that the token will never be received as they have entered a wrong email address.  Now the two factor authentication looks enabled, without the email validation step and without being able to disable it. 

Even worse, if the user logs out and tries to log in again, they will never be able to do so, as the system will be asking for a token that is being sent to a wrong email address.

claudep:
@Atterratio, @pickfire, @dchukhin, any idea for a fix?

pickfire:
We don't use email for two factor, so not affected by this, but I didn't look closely into this.

pickfire:
Ah, I tried using this and I noticed this problem. I took a look into this, my take is that confirmed should be disabled by default, and only once confirmed then it is confirmed, then only mark it as confirmed.

https://github.com/django-otp/django-otp/blob/ac020731cd30d5ef086a040fe256118be8196f69/src/django_otp/plugins/otp_email/migrations/0001_initial.py#L19

[...]

PR #752, opened 2025-03-25
Change EmailDevice default to unconfirmed until verified

MAkcanca:
Related to https://github.com/jazzband/django-two-factor-auth/issues/751

## Description
We change EmailDevice default to unconfirmed (confirmed False) because until the email is actually verified with OTP, we should not assume that flow is finished and user properly setup the OTP. This changes the default property confirmed to False, and adds a test to ensure we get a device created in a confirmed False state.

## Motivation and Context
https://github.com/jazzband/django-two-factor-auth/issues/751
https://github.com/jazzband/django-two-factor-auth/issues/562

## How Has This Been Tested?
Ran existing tests + added a new single test to confirm it creates an unconfirmed device if code is not validated. I also tested with the provided example app, starting the setup flow with email, then cancelling it without entering the code, checking from the shell/admin that email device is created with first unconfirmed, then confirmed state.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [] My change requires a change to the documentation.

### DIFF
--- a/two_factor/utils.py
+++ b/two_factor/utils.py
@@ -6,12 +6,12 @@ from django_otp import devices_for_user
 USER_DEFAULT_DEVICE_ATTR_NAME = "_default_device"
 
 
-def default_device(user):
+def default_device(user, confirmed=True):
     if not user or user.is_anonymous:
         return
     if hasattr(user, USER_DEFAULT_DEVICE_ATTR_NAME):
         return getattr(user, USER_DEFAULT_DEVICE_ATTR_NAME)
-    for device in devices_for_user(user):
+    for device in devices_for_user(user, confirmed=confirmed):
         if device.name == 'default':
             setattr(user, USER_DEFAULT_DEVICE_ATTR_NAME, device)
             return device
--- a/two_factor/views/core.py
+++ b/two_factor/views/core.py
@@ -548,6 +548,7 @@ class SetupView(RedirectURLMixin, IdempotentSessionWizardView):
         # PhoneNumberForm / YubiKeyDeviceForm / EmailForm / WebauthnDeviceValidationForm
         elif method.code in ('call', 'sms', 'yubikey', 'email', 'webauthn'):
             device = self.get_device()
+            device.confirmed = True
             device.save()
 
         django_otp.login(self.request, device)
--- a/tests/test_email.py
+++ b/tests/test_email.py
@@ -189,3 +189,22 @@ class EmailTest(UserMixin, TestCase):
         self.user.email = ""
         self.user.save()
         self.test_device_without_email()
+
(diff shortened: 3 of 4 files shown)


two_factor/utils.py
```diff
@@ -6,12 +6,12 @@
...
 
 
-def default_device(user):
+def default_device(user, confirmed=True):
     if not user or user.is_anonymous:
         return
     if hasattr(user, USER_DEFAULT_DEVICE_ATTR_NAME):
         return getattr(user, USER_DEFAULT_DEVICE_ATTR_NAME)
-    for device in devices_for_user(user):
+    for device in devices_for_user(user, confirmed=confirmed):
```
MAkcanca, line 14: As per django-otp docstring for this function, it already defaults to True

moggers87:
Tests are failing due to #754

moggers87:
@MAkcanca can up update your branch with the latest from `master`? Thanks

MAkcanca:
Merged  latest branch. Thank you for your efforts on https://github.com/jazzband/django-two-factor-auth/issues/754 with @claudep

two_factor/views/core.py
```diff
@@ -548,6 +548,7 @@ def done(self, form_list, **kwargs):
         # PhoneNumberForm / YubiKeyDeviceForm / EmailForm / WebauthnDeviceValidationForm
         elif method.code in ('call', 'sms', 'yubikey', 'email', 'webauthn'):
             device = self.get_device()
+            device.confirmed = True
```
moggers87, line 551: There are no tests that cover this change

two_factor/utils.py
```diff
@@ -6,12 +6,12 @@
 USER_DEFAULT_DEVICE_ATTR_NAME = "_default_device"
 
 
-def default_device(user):
+def default_device(user, confirmed=True):
```
moggers87, line 9: The changes here are partially covered by the test added in `tests/test_email.py` but I think we could do with some more thorough testing, e.g. test that `default_device(user)` returns only confirmed email devices and `default_device(user, None)` returns both confirmed and unconfirmed.

moggers87 (changes requested):
Just a few tests missing, otherwise looks good :+1:

two_factor/utils.py
MAkcanca: Just added 2 new tests to address this. Let me know what you think.

two_factor/views/core.py
MAkcanca: I assumed that default setup flow already covers it since it uses confirmed property as default, but added a new test to cover that scenario as well to separate the concerns. Thanks

Issue #754, opened 2025-03-26
Webathun tests randomly fail


moggers87:
## Expected Behavior

Tests pass 🤓 

## Current Behavior

A webauthn test fails, but only on Python 3.11 or 3.12. It's usually something like this:

```
  ======================================================================
  ERROR: test_webauthn_attestation_and_assertion (two_factor.plugins.webauthn.tests.test_e2e.E2ETests.test_webauthn_attestation_and_assertion)
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/home/runner/work/django-two-factor-auth/django-two-factor-auth/./two_factor/plugins/webauthn/tests/test_e2e.py", line 89, in test_webauthn_attestation_and_assertion
      self.register_authenticator()
    File "/home/runner/work/django-two-factor-auth/django-two-factor-auth/./two_factor/plugins/webauthn/tests/test_e2e.py", line 76, in register_authenticator
      self.webdriver.find_element(By.XPATH, "//button[@type='submit']").click()
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/work/django-two-factor-auth/django-two-factor-auth/.tox/py311-dj50-webauthn/lib/python3.11/site-packages/selenium/webdriver/remote/webdriver.py", line 742, in find_element
      return self.execute(Command.FIND_ELEMENT, {"using": by, "value": value})["value"]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/work/django-two-factor-auth/django-two-factor-auth/.tox/py311-dj50-webauthn/lib/python3.11/site-packages/selenium/webdriver/remote/webdriver.py", line 348, in execute
      self.error_handler.check_response(response)
    File "/home/runner/work/django-two-factor-auth/django-two-factor-auth/.tox/py311-dj50-webauthn/lib/python3.11/site-packages/selenium/webdriver/remote/errorhandler.py", line 229, in check_response
      raise exception_class(message, screen, stacktrace)
  selenium.common.exceptions.NoSuchElementException: Message: no such element: Unable to locate element: {"method":"xpath","selector":"//button[@type='submit']"}
[...]

PR #755, opened 2025-03-26
Fix webauthn test weirdness by waiting for login to finish

moggers87:
## Description

In webauthn end-to-end tests we now wait for login to finish before proceeding with 2FA setup . Without this the Selenium ends up loading the wrong URL

## Motivation and Context

See #754 

## How Has This Been Tested?

Tests now pass.

## Types of changes

## Checklist:

### DIFF
--- a/two_factor/plugins/webauthn/tests/test_e2e.py
+++ b/two_factor/plugins/webauthn/tests/test_e2e.py
@@ -71,6 +71,9 @@ class E2ETests(UserMixin, StaticLiveServerTestCase):
         button_next = self.webdriver.find_element(By.XPATH, "//button[@type='submit']")
         button_next.click()
 
+        # wait for login to complete before moving on
+        self.wait_for_url(urljoin(self.base_url, reverse("two_factor:profile")))
+
     def register_authenticator(self):
         self.webdriver.get(urljoin(self.base_url, reverse("two_factor:setup")))
         self.webdriver.find_element(By.XPATH, "//button[@type='submit']").click()


two_factor/plugins/webauthn/tests/test_e2e.py
```diff
@@ -71,6 +72,9 @@ def do_login(self):
         button_next = self.webdriver.find_element(By.XPATH, "//button[@type='submit']")
         button_next.click()
 
+        # sleep to fix #754
+        time.sleep(3)
+
     def register_authenticator(self):
         self.webdriver.get(urljoin(self.base_url, reverse("two_factor:setup")))
         self.webdriver.find_element(By.XPATH, "//button[@type='submit']").click()
```
claudep, line 80: My suggested fix would be rather to transform this `find_element()` by a `wait_for_element()`, like the lines below. Would that make sense?

claudep: Answering to myself: Would not hurt, but probably not sufficient. Anyway I would prefer using some wait at the end of the login rather than the sleep.

moggers87: No because Selenium is loading the profile page rather than the setup page so there's nothing to wait for at that point. But there is `wait_for_url` which is probably what I actually want instead of sleep!

claudep: I was just looking to `wait_for_url` :-) Yes, should be the right fix™

claudep (approved):
Thanks!

PR #558, opened 2022-09-29
fix regression from PR #500

PetrDlouhy:
Fix regression introduced in #500.

## Description
All the changes of `test_view_login.py` were wrong and they are reverted. 
We are now detecting if OTPRequiredMixin is used for `next` URL and only then we redirect to OTP setup page.
The newly introduced test in OTPRequiredMixinTest is not changed and passes.

## Motivation and Context
Fix regression described in #499

## How Has This Been Tested?
- TOX tests
- tests of Login view on my own project that were not passing after #500

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #757, opened 2025-04-11
Fix infinite login redirect on admin site with AdminSiteOTPRequiredMixin

shawnz:
When using AdminSiteOTPRequiredMixin, users who don't have 2FA set up will face an infinite login redirect, similar to the behaviour prior to #499/#500.

In #558, a change was made to fix a regression introduced by #500 by restricting when the 2FA setup login redirect will apply. But, the restriction logic only considers the use of OTPRequiredMixin. It doesn't consider the use of AdminSiteOTPRequiredMixin. Thus, users of AdminSiteOTPRequiredMixin on their admin site will still see the same broken infinite redirect behaviour from before.

This changes the logic added in #558 to also consider AdminSiteOTPRequiredMixin in addition to OTPRequiredMixin.

## Description
Update the is_otp_view logic to additionally consider views that are part of an admin site with the AdminSiteOTPRequiredMixin as OTP views.

## Motivation and Context
Fixes the infinite login redirect for admin sites, first described in #499, fixed in #500, then regressed for admin sites specifically in #558.

## How Has This Been Tested?
I applied this change to my workplace's Django 4.2.20 app and tried it out by hand to confirm that the infinite redirect is fixed for our admin site. Additionally, I copied the test from #500 which tests the redirect flow for logged out users without OTP on OTP required views but applied it to the OTPAdminSite.

Before the change:

```
FAIL: test_otp_admin_login_redirect_without_otp (tests.test_admin.OTPAdminSiteTest.test_otp_admin_login_redirect_without_otp)
----------------------------------------------------------------------
  File "/Users/shawnz/Documents/GitHub/django-two-factor-auth/.venv/lib/python3.13/site-packages/django/test/utils.py", line 456, in inner
    return func(*args, **kwargs)
  File "/Users/shawnz/Documents/GitHub/django-two-factor-auth/tests/test_admin.py", line 91, in test_otp_admin_login_redirect_without_otp
    self.assertRedirects(response, reverse('two_factor:setup'))
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 302 != 200 : Couldn't retrieve redirection page '/otp_admin/': response code was 302 (expected 200)

----------------------------------------------------------------------
Ran 165 tests in 2.713s

FAILED (failures=1, skipped=1)
```

After the change:

```
----------------------------------------------------------------------
Ran 165 tests in 2.709s

OK (skipped=1)
```

N/A

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_admin.py
+++ b/tests/test_admin.py
@@ -1,11 +1,15 @@
+from binascii import unhexlify
+
 from django.conf import settings
 from django.shortcuts import resolve_url
 from django.test import TestCase
 from django.test.utils import override_settings
+from django.urls import reverse
+from django_otp.oath import totp
 
 from two_factor.admin import patch_admin, unpatch_admin
 
-from .utils import UserMixin
+from .utils import UserMixin, method_registry
 
 
 @override_settings(ROOT_URLCONF='tests.urls_admin')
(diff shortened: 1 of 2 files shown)


moggers87 (changes requested):
Change looks good but I think it needs tests otherwise we will run into the same problem again in the future.

shawnz:
Agreed, working on them as we speak! 

Thanks for looking at this

shawnz:
@moggers87 I copied the test from #500 which validates the login and setup redirect flow but applied it to the OTPAdminSite. Let me know if you feel that's sufficient.

Thanks, Shawn

moggers87 (approved):
LGTM

Issue #469, opened 2022-01-29
Make phonenumber plugin really a plugin


claudep:
In commit 07dac39d35783, we laid the foundation for a new plugin architecture, and the PhoneNumber capability was moved to a plugin. However, there are still places in the main code that expect the phone plugin to be activated.

This is some meta issue to make the phonenumber plugin really optional.

sevdog:
There is also an import from phonedevice plugin in `views.core`:

https://github.com/jazzband/django-two-factor-auth/blob/22aaab5c702d44d26ee8bf83c0190aca997e0331/two_factor/views/core.py#L38

Which breaks when importing anything from `views` package if `phonenumbers` is not installed:

```
  File ".virtualenvs/lib/python3.9/site-packages/two_factor/views/__init__.py", line 1, in <module>
    from .core import (
  File ".virtualenvs/lib/python3.9/site-packages/two_factor/views/core.py", line 38, in <module>
    from two_factor.plugins.phonenumber.utils import get_available_phone_methods
  File ".virtualenvs/lib/python3.9/site-packages/two_factor/plugins/phonenumber/utils.py", line 3, in <module>
    import phonenumbers
ModuleNotFoundError: No module named 'phonenumbers'
```

claudep:
Sure, probably ` get_available_phone_methods` should be moved (back?) in `two_factor.utils`?

sevdog:
> Sure, probably ` get_available_phone_methods` should be moved (back?) in `two_factor.utils`?

Since that method does not rely on other code in the phonenumber plugin it could work.

But IMO it would be better to find a way to have the `SetupCompleteView` to get extra context using the registry.

Also, now that I was looking for that method (and other imports from that plugin), there is an import from phone plugin also in `two_factor.views.profile` (which is also imported in `__init__` file):

https://github.com/jazzband/django-two-factor-auth/blob/83ce8005f76d37a2b00d5ccda03819619a9ee669/two_factor/views/profile.py#L10-L12

To have a nice plugin system also the `ProfileView` should get its data from the registry. Otherwive there it would just be hardcoded and not plugin-based.

claudep:
Please feel free to suggest a PR with your ideas!

swiss-knight:
Hello,

may I link [this SO thread](https://stackoverflow.com/questions/75570862/django-two-factor-authphonenumbers-got-a-redundant-migration-psycopg2-error) without necessarily opening a new issue as it _may_ be related?

[...]

PR #765, opened 2025-05-31
Avoid hard dependency on phonenumbers in migrations

claudep:
Refs #469.


moggers87 (approved):
Looks good, though I think we still have a few more places where phonenumber is imported.

PR #792, opened 2026-05-06
fix #652: pick most-recently-used 2FA device when no device is named 'default'

### LINKED ISSUE #652: Enabling 2FA first time for user does not work if TOTP device name is not 'default'
Hello everyone,
I am not sure if this has been fixed on a later version but I haven t found anything about it so I thought to post my issue. The bug occurs when for the first time you assign a TOTP device to a user. If the name of the device is not 'default' the user does not get flagged as 2FA enabled and hence the 2FA is not enabled. If instead you change the name of the TOTP device to 'default' or you create a new TOTP device named 'default' then the 2FA gets enabled for the user.

## Expected Behavior
Login without 2FA. Create a TOTP device with any name I want. Logout. Login, 2FA required.

## Current Behavior
1) Login without 2FA. Create a TOTP device with any name I want. Logout. Login, 2FA is not requried.
2) Login without 2FA. Create a TOTP device with 'default' name. Logout. Login, 2FA is required.

1. Login with user that has no 2FA enabled and no TOTP Devices assigned
2. Create TOTP Device with name different from 'default'
3. Logout
4. Login again

## Context
I would like to be able to name TOTP devices to know where they are used (i.e. authenticator, 1password, ...)

## Your Environment
* Browser and version: tested on many always happens
* Python version: 
* Django version: 3.1.14
* django-otp version: 1.1.6

If this has been fixed in latest releases would be cool if someone could let me know.
Thanks everybody

PetrDlouhy:
## Description
`default_device(user)` matched only `device.name == "default"`. Users naming their device anything else (e.g. `"YubiKey"`, `"1Password"`) were treated as having no 2FA at all.
New selection order: a device named `"default"` (back-compat), then the most-recently-used non-backup device, then the lowest `persistent_id` as tie-breaker. Backup devices (`StaticDevice` and any device named `"backup"`, the same convention as `backup_phones()`) are excluded.
Adds `TWO_FACTOR_DEFAULT_DEVICE_PICKER` setting — dotted path to a callable `(devices) -> Device | None` — for projects that want a different policy without monkey-patching. Public helper `primary_device_candidates(devices)` is exported so custom pickers can reuse the backup-exclusion logic.
If you'd rather keep the legacy default (only `name="default"`) and ship MRU as an opt-in built-in picker, happy to refactor.
## Motivation and Context
Fixes #652.
## How Has This Been Tested?
Tox (`py39-dj42-{normal,yubikey,custom_user,webauthn}`), unit tests added for every branch of the new policy and the picker hook.
`test_webauthn_attestation_and_assertion` was renaming the device to `'not default anymore'` to bypass the setup wizard's `if default_device(user): redirect` guard. With the new policy that workaround stops working, so the rename target is now `'backup'` (using the same exclusion convention exposed by `primary_device_candidates`). The credential ID stays in the DB, so the duplicate-detection path being tested still fires.
## Types of changes
- [ ] Breaking change
## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,22 @@
 ## Unreleased
 ### Fixed
 - Compile translation files for Azerbaijani and Serbian, forgotten in 1.18.1.
+- `default_device()` now returns the user's most-recently-used non-backup
+  device when no device is named `"default"`, instead of returning `None`.
+  This fixes [#652](https://github.com/jazzband/django-two-factor-auth/issues/652):
+  users naming their first device anything other than `"default"` (e.g.
+  `"YubiKey"`, `"1Password"`) were never flagged as having 2FA enabled.
+  Backward-compatible: deployments that rely on a device named `"default"`
+  see no change.
+
+### Added
+- `TWO_FACTOR_DEFAULT_DEVICE_PICKER` setting: dotted path to a callable
+  that overrides the built-in primary-device selection policy. Lets
+  projects implement their own policies (e.g. "always prefer WebAuthn",
+  "alphabetical by name", "specific device class first") without
+  monkey-patching `default_device`. New public helper
+  `two_factor.utils.primary_device_candidates(devices)` is exported for
+  custom pickers that want to reuse the built-in backup-exclusion logic.
 
 ## 1.18.1
 ### Added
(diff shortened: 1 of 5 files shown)


moggers87 (approved):
Nice clean code with lots of comments 👍

PR #797, opened 2026-08-04
Revert "fix #652: pick most-recently-used 2FA device when no device is named 'default'"

### LINKED ISSUE #652: Enabling 2FA first time for user does not work if TOTP device name is not 'default'

moggers87:
Reverts jazzband/django-two-factor-auth#792

I believe I made a mistake mergeing #792 

I didn't it notice at the time but #792 was AI generated and realising that caused me to re-review the PR. As well as the absolutely huge (and often redundant) comments there's at least one test (`test_keeps_order_of_input`) that doesn't properly do what it says it does which makes me doubt the entire PR. It's a small oversight but AIs don't make mistakes like humans do - it didn't just "forget" to properly test the order of input vs output - and it makes me wonder what else is wrong.

As that commit has *not* been released yet I think we're safe to revert and then reassess. The main bulk of the code might be fine but I no longer feel confident that I properly reviewed that code the first time around.


claudep (approved):
Considering my "AI-reluctance", I can only approve :rofl:

PR #801, opened 2026-08-24
Setup wizard: don't reset state on a GET while in progress

### LINKED ISSUE #767: BUG: TOTP validation fails on all browsers except Safari
## Expected Behavior
TOTP token validation should work consistently across all browsers after scanning the QR code and entering the 6-digit token from the authenticator app.

## Current Behavior
- Safari: Token validation always succeeds ✅
- Chrome/Firefox/Edge: Token validation always fails with "Entered token is not valid" ❌

The TOTP comparison in `forms.py` never evaluates to True on non-Safari browsers:
```python
if totp(key, self.step, t0, self.digits, self.drift + offset) == token:
    # This is never True on Chrome/Firefox/Edge
```

## Possible Solution
Since it's browser-specific, it might be related to:
- How different browsers handle form data encoding
- Session/cookie handling differences
- other ... ?

1. Setup django-two-factor-auth following the standard documentation
2. Login and start 2FA setup process
3. Scan QR code with any authenticator app
4. Enter the 6-digit token
5. Submit form
6. Result: Works on Safari, fails on any other browser

## Context
I'm implementing 2FA for a Django application. Everything works perfectly on Safari but fails consistently on all other browsers. This is blocking deployment as users can't enable 2FA unless they use Safari.

## Your Environment
- Browser and version: Safari (works), Chrome/Firefox/Edge latest versions (fail)
- Python version: 3.12.11
- Django version: 4.0.5
- django-otp version: (included with django-two-factor-auth)
- django-two-factor-auth version: 1.16.0

aroh3006:
Fixes #767.

SetupView.get() falls through to WizardView.get(), which resets the wizard on every GET, including the just-generated TOTP secret. If a browser sends a background GET to the setup URL while the form is still open (dev tools, prefetch, a second tab), the secret is silently replaced. The token the user later submits, generated from the secret shown in the QR code, then fails validation against a different key, and setup can never complete.

SetupView.get() now only resets when no wizard is in progress. It re-renders the current step from existing state otherwise.

Added a test that posts through to the generator step, fires a GET at the same URL to simulate the stray request, then confirms the originally shown key still validates. It fails on master (200 instead of a redirect) and passes with the fix.

Ran the full test suite (162 passed) plus the webauthn plugin tests, ruff, and isort.

### DIFF
--- a/two_factor/views/core.py
+++ b/two_factor/views/core.py
@@ -482,6 +482,14 @@ class SetupView(DeviceContextDataMixin, RedirectURLMixin, IdempotentSessionWizar
         """
         if default_device(self.request.user):
             return redirect(self.get_success_url())
+        if self.storage.current_step is not None:
+            # A wizard is already in progress. WizardView.get() always
+            # resets the wizard, but some browsers send a background GET
+            # to the current page while the user is mid-setup (see issue
+            # #767). Resetting here would silently drop the TOTP secret
+            # already shown to the user, so every later token submission
+            # would fail against a key that no longer matches it.
+            return self.render(self.get_form())
         return super().get(request, *args, **kwargs)
 
     def get_form(self, step=None, **kwargs):
--- a/tests/test_views_setup.py
+++ b/tests/test_views_setup.py
@@ -67,6 +67,30 @@ class SetupTest(UserMixin, TestCase):
         self.assertRedirects(response, reverse('two_factor:setup_complete'))
         self.assertEqual(1, self.user.totpdevice_set.count())
 
+    @method_registry(['generator'])
+    def test_setup_generator_key_survives_stray_get(self):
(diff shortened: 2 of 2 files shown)


two_factor/views/core.py
```diff
@@ -482,6 +482,14 @@ def get(self, request, *args, **kwargs):
         """
         if default_device(self.request.user):
             return redirect(self.get_success_url())
+        if self.storage.current_step is not None:
+            # A wizard is already in progress. WizardView.get() always
+            # resets the wizard, but some browsers send a background GET
```
moggers87, line 487: Name and shame them! 😸

moggers87 (commented):
Code-wise this is fine, but I'm withholding approval because I'm not sure it's the right thing to do. I'll add a comment in #767 to explain my current thinking.

PR #804, opened 2026-09-12
Allow webauthn 3.x

ghost:
Currently `webauthn>=2.0,<2.99` pins the WebAuthn extra to 2.x. Since webauthn 3.0.0 requires `cbor2>=6.1.2`, this also holds back cbor2.

As far as I can tell there is no incompatibility behind the cap — 3.0.0's release notes only list additions (PQC/ML-DSA algorithms, EdDSA preference, stricter CBOR duplicate-key rejection), and none of the helpers this package imports from `webauthn`, `webauthn.helpers` or `webauthn.helpers.structs` changed.

Test results on `a718bc7`, Python 3.14.6, Django 5.2.17:

| webauthn / cbor2 | result |
|---|---|
| 2.8.0 / 5.9.0 | 169 tests OK (8 skipped, all YubiKey) |
| 3.0.0 / 6.1.4 | 169 tests OK (8 skipped) |

`test_webauthn_attestation_and_assertion` (Selenium + Chrome virtual authenticator) ran in both — not skipped — so registration and assertion are covered.

I normalised the cap style to `<4` rather than keeping the `<3.99` pattern used by the other extras — happy to change it back if you prefer consistency.


claudep (approved):
Thanks for testing this, I think we should go ahead with this if current tests are passing.

claudep:
@moggers87 Should we reopen a new PR with this content (looks like the proposing user was deleted)?

